PR #121, opened 2026-04-08
feat(config): add custom OpenAI-compatible provider [PR A of 3]

Anduin9527:
## Summary
  - add a single custom OpenAI-compatible provider for executor and reviewer
  - fetch custom model choices dynamically from /models during setup, /model, and /reviewer
  - fix reviewer disable state and executor credential/runtime refresh logic for provider switches

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # ARIS-Code Changelog
 
+## Unreleased
+
+- **Custom OpenAI-compatible provider**: Added a single `custom` provider for both Executor and Reviewer.
+- **Dynamic model discovery**: Custom setup, `/model`, and `/reviewer` now fetch model ids from `/models` instead of asking users to type them manually.
+- **Custom reviewer routing**: `LlmReview` now respects `ARIS_REVIEWER_PROVIDER=custom` and sends requests directly to the configured endpoint.
+
 ## v0.3.5 (2026-04-08)
 
 - **New**: **Research Wiki** — persistent research knowledge base with papers, ideas, experiments, claims, and typed relationship graph. Python helper with auto-fallback to direct LLM execution.
--- a/README.md
+++ b/README.md
@@ -88,14 +88,16 @@ The first time you run `aris`, an interactive setup wizard launches automaticall
     Google Gemini
     Zhipu GLM
     MiniMax
-Enter API Key: sk-...
+    Custom OpenAI-compatible
+Enter API Key / Base URL, then choose a model from /models
 
 [2/3] Choose Reviewer provider (adversarial LLM)
   > OpenAI GPT
     Google Gemini
     Zhipu GLM
     MiniMax
-Enter API Key: sk-...
+    Custom OpenAI-compatible
+Enter API Key / Base URL, then choose a model from /models
 
 [3/3] Choose language preference
     中文 (CN)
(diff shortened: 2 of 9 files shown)


wanshuiyin:
Hi @Anduin9527 — thanks for this contribution. The direction is right (ARIS-CLI's provider handling has clear duplication across LlmReview / CLI executor / subagent runtime, and we agree custom OpenAI-compatible providers + dynamic `/models` is where this should go).

We've just shipped **v0.4.5** today, which makes 3 things relevant:

1. The default model + setup UX has moved on significantly since this PR was opened (Claude Opus 4.7 + GPT-5.5 + DeepSeek + Xiaomi + Qwen + Doubao + `reasoning_effort=xhigh`).
2. v0.4.4's `resolve_reviewer_model()` is now a hard regression-test contract.
3. PR #186 / #171 / #216-partial all landed, so the surrounding code has changed.

**Your current diff (+1638/-386 across the same files we modified) would substantially overwrite all of this, and we can't tell from the current state whether the regressions are preserved.** Asking for a rework before we can merge.

### What we'd like you to do

#### 1. Rebase onto current `aris-code` (post-v0.4.5)

Pick up the v0.4.5 work. After rebase the actual delta should be clearer.

#### 2. Split into 3 PRs

- **PR A** — Provider schema/catalog: data model for "custom OpenAI-compatible provider" entry (config schema + setup wizard menu entry). Smallest, safest.
- **PR B** — Executor dynamic `/models` discovery: `openai_compat.rs` for fetching model list, hooked into `/setup`, `/model`, `/reviewer`. **Must be mockable for offline testing — CI must not hit real `api.openai.com`.**
- **PR C** — Reviewer + executor shared routing: factor a provider descriptor used by both `OpenAIRuntimeClient` and `LlmReview` (currently the two have independent OpenAI/Anthropic routing — see `aris-cli/openai_executor.rs:87` vs `tools/lib.rs:3174`).

#### 3. Preserve these non-negotiable v0.4.x invariants

Tests that must keep passing (please run `cargo test -p aris-cli -p tools`):

**Setup** (`aris-cli/config.rs`):
- Choice/provider separation — menu choice index ≠ provider string. Gemini/GLM/MiniMax/Kimi/Xiaomi/Qwen/Doubao all serialize as `provider="openai"`; only the display label differs.
- Custom URL preservation — repeated `/setup` runs must not overwrite user-set custom base URLs.
- Reviewer stale state clearing — switching reviewer provider must clear stale key/base_url so the old reviewer doesn't leak.

**Reviewer routing** (`tools/lib.rs`):
- `resolve_reviewer_model()` provider consistency rule (override only honored if `provider==configured` AND key exists).
- All three regression tests must pass: missing-key fallback, provider-mismatch fallback, stale `ARIS_REVIEWER_PROVIDER` does not block legitimate override.

**Anthropic betas** (`api/client.rs`):
- `read_send_betas()` must continue to read `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS`; setup must continue to auto-set it for custom Anthropic base URLs.

**reasoning_effort** (v0.4.5 new):
- New executor + reviewer path attaches `reasoning_effort` for `gpt-5.5` / `o1*` / `o3*` / `o4*` / `*-reasoner` / `*-thinking`. Please don't strip this when refactoring the request body builder.

#### 4. Keep `openai_compat.rs` as a shared module

Your new file is the right direction. Factor it so the **same provider descriptor** is used by both `OpenAIRuntimeClient` (streaming + tool-call runtime) and `LlmReview` (blocking non-stream reviewer). Don't add a third independent routing path.

## Timeline

- v0.4.5 just shipped.
- v0.4.6 is for native-tls + DashScope (#159).
- Your reworked PRs would target a **`v0.5.0-alpha/provider-abstraction`** branch (we'll open it after native-tls lands).

Happy to pair on the rebase or split if helpful — ping us or open an issue. Thanks for the patience with the rework ask.

— maintainer team

wanshuiyin:
Hi @Anduin9527 — closing this PR since you reworked it into #221 (PR A) and #222 (PR B), both of which have now been cherry-picked into `aris-code` (commits `e737d69` and `784140c`) for the v0.4.6 release.

See #221 / #222 for the full landing notes. Thanks for the patient split.

— maintainer team

PR #126, opened 2026-04-08
feat: integrate deepxiv into aris workflows

DreamEnding:
**Summary**
- add a thin `tools/deepxiv_fetch.py` adapter around the installed `deepxiv` CLI
- add standalone `deepxiv` markdown skills for both base ARIS and `skills-codex`
- integrate `deepxiv` into `research-lit` as an explicit opt-in source
- keep existing defaults unchanged: `sources: all` still does not include `deepxiv`
- document optional installation and usage in README files

**Why**
- ARIS already supports arXiv and Semantic Scholar, but lacks a progressive-reading paper source
- DeepXiv adds layered retrieval (`search -> brief -> head -> section`) without replacing existing flows
- this keeps the integration non-breaking and optional


wanshuiyin:
Thanks @DreamEnding for the DeepXiv integration! The overall design is solid — opt-in source, graceful degradation, clean skill + adapter + tests structure.

However, we found a few issues that need fixing before merge:

### 1. 🔴 `paper-head` will crash at runtime
`paper-head` is routed through `run_cli_json()` in `tools/deepxiv_fetch.py`, but the CLI output for this subcommand is plain text, not JSON. This will raise a parse error. Either:
- Add `--format json` if the CLI supports it for `paper-head`, or
- Switch that branch to text parsing instead of `run_cli_json()`

### 2. 🔴 JSON flag inconsistency
The PR uses three different spellings across docs, adapter, and tests:
- `--output json` (in tests)
- `--format json` (in adapter)
- `--json` (in skill docs)

Please verify against the actual `deepxiv` CLI and use one consistent flag everywhere.

### 3. 🟡 Missing test coverage
No tests for `paper-head`, `trending`, `wsearch`, `sc` branches, or the failure path when CLI stdout is not valid JSON. The `paper-head` bug would have been caught with a test.

### 4. 🟡 Unrelated `.gitignore` change
The `.gitignore` addition for `docs/superpowers/` is unrelated to DeepXiv — please drop this hunk or justify it separately.

---

Once these are fixed, we're happy to merge! The staged retrieval design (search → brief → head → section) is a great fit for ARIS's research-wiki pipeline.

DreamEnding:
Thanks again for the detailed review! I've addressed all the issues:

1. **`paper-head` crash:** Fixed. I added the `--format json` flag to the `paper --head` command. It now outputs valid JSON and works perfectly with `run_cli_json()`.
2. **JSON flag inconsistency:** Fixed. I verified the flags against the actual `deepxiv` CLI. It turns out the CLI uses slightly different flags depending on the subcommand. I've updated the codebase to accurately reflect the actual CLI behavior:
   - `paper ... --head` uses `--format json`
   - `trending`, `wsearch`, and `sc` use `--output json`
   The adapter, tests, and skill docs are now completely consistent with these verified commands.
3. **Missing test coverage:** Added. I've introduced comprehensive tests for the `paper-head`, `trending`, `wsearch`, and `sc` branches, including the failure path when CLI stdout is not valid JSON.
4. **Unrelated `.gitignore` change:** Reverted. I have dropped the `docs/superpowers/` addition from this PR.

I have also thoroughly tested these changes locally, and everything is running smoothly. 

Just to add a brief background: DeepXiv is a free, open-source literature search project specifically optimized for AI-driven fuzzy literature searching. Adapting and integrating DeepXiv into the ARIS workflows is a perfect match, as it significantly empowers the research-wiki pipeline to handle unstructured queries and retrieve academic context more effectively.

PR #164, opened 2026-04-21
feat: add /arxivsub-skill — arXivSub API paper search

arxivsub:
## Summary                                                                    
          
- Adds `/arxivsub-skill`: searches arXiv and major AI/CV conferences (CVPR,   
ICCV, ICLR, ICML, NeurIPS, AAAI, MICCAI) via the arXivSub API                 
- arXivSub maintains a managed database of papers, making it easy to find the
newest research quickly                                                       
- Requires `ARXIVSUB_SKILL_KEY` — get your API key at                         
https://arxivsub.comfyai.app/
                                   
## Test plan                                                                  
                                                                           
- [ ] Set `ARXIVSUB_SKILL_KEY` (refer to [arxivsub-skill](https://github.com/arxivsub/arxivsub-skill)) and run `/arxivsub-skill` in Claude Code       
- [ ] Verify search results are returned for an arXiv query                   
- [ ] Verify conference-specific search (e.g. `locations: NeurIPS`) works

### DIFF
--- a/README.md
+++ b/README.md
@@ -1110,6 +1110,7 @@ export OPENAI_API_KEY="your-key"
 |-------|-------------|:---:|
 | 📄 [`arxiv`](skills/arxiv/SKILL.md) | Search, download, and summarize arXiv papers. Standalone or `/research-lit` supplement | No |
 | 🔎 [`semantic-scholar`](skills/semantic-scholar/SKILL.md) | Search published venue papers (IEEE, ACM, Springer) via Semantic Scholar API. Citation counts, venue metadata, TLDR | No |
+| 🗂️ [`arxivsub-skill`](skills/arxivsub-skill/SKILL.md) | Search arXiv and major AI/CV conferences (CVPR, ICCV, ICLR, ICML, NeurIPS, AAAI, MICCAI) via arXivSub API. Bundled Python scripts (`search.py`, `fetch.py`). Requires `ARXIVSUB_SKILL_KEY` | Yes (`ARXIVSUB_SKILL_KEY`) |
 | 📚 [`deepxiv`](skills/deepxiv/SKILL.md) | Progressive paper retrieval via DeepXiv CLI: search, brief, section map, section reads, trending, web search | Yes (`pip install deepxiv-sdk`) |
 | 🔎 [`exa-search`](skills/exa-search/SKILL.md) | AI-powered broad web search via Exa: blogs, docs, news, companies, research papers with content extraction (highlights, text, summaries) | Yes (`pip install exa-py`) |
 | 📝 [`alphaxiv`](skills/alphaxiv/SKILL.md) | Quick single-paper lookup via [AlphaXiv](https://alphaxiv.org) LLM-optimized summaries. Three-tier fallback: overview → full markdown → LaTeX source | No |
--- a/skills/arxivsub-skill/SKILL.md
+++ b/skills/arxivsub-skill/SKILL.md
@@ -1,0 +1,126 @@
+---
+name: arxivsub-skill
+description: >
+  Academic paper search assistant that queries the arXivSub API to find the latest research papers
+  from arXiv and major AI/CV conferences (CVPR, ICCV, ICLR, ICML, NeurIPS, AAAI, MICCAI).
+  Use this skill whenever the user wants to search for academic papers, find recent research,
+  look up conference publications, discover what's new in a research area, or explore literature
+  on any AI/ML/CV topic. Trigger even for casual phrasing like "find papers on X", "what are
+  the latest papers about Y", "search arXiv for Z", or "any recent work on W".
+---
+
+# arxivsub-skill
+
+Search academic papers via the arXivSub API.
+
+Two bundled scripts handle all processing: `scripts/search.py` (API fetch + parse) and `scripts/fetch.py` (index/detail retrieval).
+
+---
+
+## Language Rule
+
+**Always respond in the same language the user is using.**
+
+---
+
+## Step 0: Authentication
+
+The API key is read automatically by `search.py` from the environment — never ask the user for it and never pass it as a CLI argument.
+
+If `search.py` exits with a non-zero code mentioning `ARXIVSUB_SKILL_KEY`, tell the user (in their language) to set it up via **one** of these methods:
+
+1. Export as a shell environment variable (add to `~/.zshrc` or `~/.bashrc` for persistence):
+   ```
+   export ARXIVSUB_SKILL_KEY=your_key_here
+   ```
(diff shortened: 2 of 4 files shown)


wanshuiyin (changes requested):
Thanks for the contribution — the **pre-parsed structured-summary schema** (`what_about / innovations / techniques / datasets / results / limitations`) is a genuine value-add over `/arxiv` and `/semantic-scholar`, and bundling the AI/CV conference index is useful. Happy to merge once a few items are addressed.

ARIS already has 7 paper-search skills (`/arxiv`, `/semantic-scholar`, `/deepxiv`, `/alphaxiv`, `/exa-search`, `/gemini-search`, `/openalex`) plus the orchestrator `/research-lit`. Vendor-specific integrations are welcome (we already have Exa, MiniMax, Gemini), but they should be **opt-in additions**, not defaults that override existing single-source skills or the orchestrator.

## Blockers

### 1. Trigger description is too greedy

The current description asks the skill to fire on *any* paper-search intent, including `"search arXiv for Z"` and `"find papers on X"`. That collides directly with `/arxiv`, `/semantic-scholar`, and the canonical `/research-lit` workflow.

Please reframe as **opt-in**:

```yaml
description: "Search arXiv + 7 AI/CV conferences (CVPR, ICCV, ICLR, ICML, NeurIPS, AAAI, MICCAI) via the arXivSub API with pre-parsed summaries (what_about / innovations / techniques / datasets / results / limitations). Use only when the user explicitly names arxivsub/arxivsub-skill, or asks for the arXivSub bundled AI/CV index / pre-parsed summaries and ARXIVSUB_SKILL_KEY is configured. Do NOT trigger for generic paper-search requests — those should go to /research-lit, /arxiv, or /semantic-scholar."
```

If `/research-lit` integration would be useful, please add it explicitly as an opt-in source in a follow-up (e.g., `— sources: arxivsub`) rather than via greedy triggering here.

### 2. Third-party / paid-service disclosure is missing

The SKILL.md only says "Requires `ARXIVSUB_SKILL_KEY`" and the README says "Yes (`ARXIVSUB_SKILL_KEY`)". A first-time user could reasonably assume this is a free signup. The skill also returns a `quota_remaining` field but never explains what that means.

Please add explicit disclosure, e.g., at the top of `## Step 0: Authentication`:

> **Note:** arXivSub is a third-party hosted service. Queries are sent to `qtevnmgyobilaanrzidq.supabase.co`. Usage is subject to arXivSub's quota / pricing / terms of service. See https://arxivsub.comfyai.app/ for account / pricing details.

And update the README table entry from `Yes (ARXIVSUB_SKILL_KEY)` to something like `Third-party API key (account/subscription may apply)`.

### 3. Frontmatter doesn't match ARIS conventions

Compare to `/arxiv`, `/semantic-scholar`, `/exa-search`: all use single-line `description` (not YAML block scalar `>`), and all declare `argument-hint` + `allowed-tools`.

Please:

- Replace `description: >` block scalar with a single-line quoted string (one paragraph).
- Add `argument-hint: "<query> [— locations: arxiv,CVPR,...] [— limit: N] [— arxiv-days: N] [— conf-years: 2024,2025]"`
- Add `allowed-tools: Bash(*), Read, Write` (or the actual minimum set the scripts need)

This keeps the skill index renderer / router parsing consistent.

## High-risk

### 4. API URL hard-coded to a Supabase project URL

`scripts/search.py:21` has:

```python
API_URL = "https://qtevnmgyobilaanrzidq.supabase.co/functions/v1/agent-skills-gateway"
```

Tying a public-facing skill to a specific Supabase project ID is fragile — if arXivSub migrates infrastructure, every installed copy breaks until the skill is updated. Please:

- Keep the current URL as the default.
- Read `ARXIVSUB_API_URL` env override before falling back to the default.
- Document the override in SKILL.md (one-line note).

A stable public domain (e.g., `api.arxivsub.com`) would be even better long-term, but env override unblocks merge.

### 5. `<SEG>`-delimited schema parsing is fragile

`_parse_summary` in `scripts/search.py:54-84` splits on `<SEG>` and decides field assignment with magic numbers (10 vs 11 segments). For fewer than 10 segments it silently truncates `summary_content[:300]` and zeroes out the other five structured fields; for extra or otherwise-unexpected segment counts it still proceeds by positional index, which can hide schema drift silently. Two problems:

1. arXivSub controls both the API and this client — the API should return a JSON object (`{what_about, innovations, ...}`) rather than a delimited string the client guesses at.
2. The silent fallback hides backend schema drift from users.

Minimum acceptable fix:

- Detect a structured response (e.g., `"structured_summary"` field as a JSON object) and use it directly.
- Keep `<SEG>` parsing as a legacy fallback **only**.
- On unexpected segment count, emit a clear stderr warning ("arXivSub returned unexpected summary schema; some structured fields may be missing") rather than silently truncating.

Long-term: ship a versioned schema on the API side (`schema_version: "v2"`) so the client can pin behaviour.

## Medium

### 6. Retry behaviour documented but not implemented

SKILL.md `## Key Rules` says:

> Retryable (network timeout, temporary server error): inform the user and offer to retry automatically

But `search.py` has no retry / backoff logic. Either:

- Add a small retry (2 attempts with 1s + 2s backoff) for `urllib.error.URLError` and HTTP 5xx, **or**
- Change the SKILL.md wording to "report retryable errors to the user" so the doc matches behaviour.

### 7. `.env` lookup is narrow

`load_api_key()` only reads `.env` from `os.getcwd()`. Skills are typically run from inside a project directory, so this is acceptable, but consider also checking `~/.config/arxivsub/.env` or the skill directory. Not a blocker.

## Scope / framing

The skill currently advertises "major AI/CV conferences" and then claims "any AI/ML/CV topic". The covered conferences are CVPR / ICCV / ICLR / ICML / NeurIPS / AAAI / MICCAI — that's an AI/CV bias and doesn't include NLP (ACL, EMNLP, NAACL), data mining (KDD, WWW), graphics (SIGGRAPH), or robotics (RSS, CoRL, ICRA). Not a blocker, but please be explicit about scope, e.g.,:

> Currently indexes: CVPR, ICCV, ICLR, ICML, NeurIPS, AAAI, MICCAI. NLP / graphics / robotics / data-mining venues are not yet covered — for those, use `/semantic-scholar` or `/research-lit`.

## Summary

This is a real differentiator (pre-parsed structured summaries + conference bundle) and it should land. The blockers are about positioning (don't claim the default `/arxiv`-or-`/research-lit` route), disclosure (third-party service), and routine code hygiene (URL override, schema robustness). The high-risk items are concrete, mechanical fixes — not architectural changes.

Once the 3 blockers and 2 high-risk items are addressed, I'd be happy to merge. The medium / scope items are nice-to-have follow-ups but not required for this PR.

PR #166, opened 2026-04-22
[codex] Add paper-illustration-image2 bridge for native Codex image generation

kbr19-thu:
## What changed

This PR adds a new `paper-illustration-image2` skill without modifying the existing `paper-illustration` workflow.

The new path is designed for Claude-first workflows that want to trigger native Codex image generation via a local MCP bridge instead of relying on external image APIs.

## Highlights

- add a new `skills/paper-illustration-image2/` skill modeled after the existing illustration workflow
- add a new `mcp-servers/codex-image2/` MCP bridge that calls the local Codex app-server image-generation path
- reject shell / Python bitmap fallbacks and require a real native `imageGeneration` event before reporting success
- add unit coverage for the bridge behavior in `tests/test_codex_image2_server.py`

## Scope note

This path targets native raster image generation through Codex. It does not claim direct SVG/vector export from GPT Image, since the official public docs currently describe raster output formats rather than SVG output.

## Why

`paper-illustration` is currently centered around Gemini image generation. For Codex users, native image generation can already be available locally, so this PR adds a separate skill that keeps the original workflow intact while enabling a Codex-native path.

## Validation

- `python3 -m unittest tests/test_codex_image2_server.py`
- `python3 -m py_compile mcp-servers/codex-image2/server.py`
- `git diff --check`
- live smoke test through the MCP bridge with native image generation
- semi-cold-start verification showing the bridge works without conversation context and mainly depends on local Codex auth state

### DIFF
--- a/mcp-servers/codex-image2/README.md
+++ b/mcp-servers/codex-image2/README.md
@@ -8,7 +8,7 @@ This bridge is intentionally narrow:
 - it accepts **text prompts**
 - it waits for a native `imageGeneration` event from Codex
 - it rejects runs that fall back to shell/Python-based image creation
-- it copies the generated image into your requested workspace path
+- it copies the generated image only into `cwd/figures/ai_generated/`
 
 ## Requirements
 
@@ -28,9 +28,11 @@ claude mcp add codex-image2 -s user -- python3 ~/.claude/mcp-servers/codex-image
 
 ## Exposed Tools
 
-- `generate`
 - `generate_start`
 - `generate_status`
+
+The bridge is intentionally **async-first**. Call `generate_start`, then poll
+`generate_status`.
 
 ## Smoke Test
 
@@ -48,8 +50,16 @@ Generate a clean blue circle on white background.
 
 Then poll `mcp__codex-image2__generate_status` until `done=true`.
 
+`outputPath` must stay under `figures/ai_generated/` inside the provided `cwd`.
+The server rejects writes outside that workspace root.
+
 ## Current Limitations
 
 - The bridge uses the current Codex app-server path and is **experimental**
 - It is optimized for **one-shot generation** rather than conversational thread replay
 - Reference images are passed as path hints in the prompt; if Codex exposes local image viewing in that session, it may inspect them
+
+## Logging
+
+- Debug logging is opt-in via `CODEX_IMAGE2_DEBUG_LOG=/path/to/debug.log`
+- Raw Codex stdout/stderr run logs are off by default; enable only when needed with `CODEX_IMAGE2_SAVE_RUN_LOGS=1`
--- a/mcp-servers/codex-image2/README.md
+++ b/mcp-servers/codex-image2/README.md
@@ -1,0 +1,55 @@
+# codex-image2 MCP Bridge
+
+Experimental bridge that lets **Claude Code** ask the local **Codex desktop app**
+to generate an image through Codex's native app-server image-generation path.
+
+This bridge is intentionally narrow:
+
+- it accepts **text prompts**
+- it waits for a native `imageGeneration` event from Codex
+- it rejects runs that fall back to shell/Python-based image creation
+- it copies the generated image into your requested workspace path
+
+## Requirements
+
+- Codex desktop app installed and signed in
+- `codex` CLI available on `PATH`
+- `codex debug app-server send-message-v2 "ping"` works on your machine
+
+## Install for Claude Code
+
+```bash
+mkdir -p ~/.claude/mcp-servers/codex-image2
+cp mcp-servers/codex-image2/server.py ~/.claude/mcp-servers/codex-image2/server.py
+chmod +x ~/.claude/mcp-servers/codex-image2/server.py
+
+claude mcp add codex-image2 -s user -- python3 ~/.claude/mcp-servers/codex-image2/server.py
+```
+
+## Exposed Tools
+
+- `generate`
+- `generate_start`
+- `generate_status`
+
+## Smoke Test
+
+After registering the server in Claude Code, ask Claude to call:
+
+```text
+mcp__codex-image2__generate_start
+```
+
+with a prompt like:
+
+```text
+Generate a clean blue circle on white background.
+```
+
+Then poll `mcp__codex-image2__generate_status` until `done=true`.
+
+## Current Limitations
+
+- The bridge uses the current Codex app-server path and is **experimental**
+- It is optimized for **one-shot generation** rather than conversational thread replay
+- Reference images are passed as path hints in the prompt; if Codex exposes local image viewing in that session, it may inspect them
--- a/mcp-servers/codex-image2/server.py
+++ b/mcp-servers/codex-image2/server.py
@@ -18,7 +18,7 @@ import sys
 import time
 import traceback
 import uuid
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from pathlib import Path
 from typing import Any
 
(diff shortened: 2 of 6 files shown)


kbr19-thu:
Closing for now while we re-check whether GPT Image can provide vector/SVG output directly.

wanshuiyin (changes requested):
Hi @kbr19-thu — thanks for turning this around so quickly, and for doing it at this level of rigor. The bridge is intentionally narrow, the README is candid about the experimental surface, and the native-tool guardrail is implemented in code rather than only promised in prose. I also appreciate that you kept this as a parallel path next to `paper-illustration` instead of mutating the existing Gemini workflow in place.

## ✅ What works well

- The "native image only" rule is real enforcement, not just a README claim: the bridge prompt forbids shell/Python fallbacks and the runtime rejects any parsed `commandExecution` item instead of silently accepting it (`mcp-servers/codex-image2/server.py:188-214,371-375`; `mcp-servers/codex-image2/README.md:8-12`).
- The async job shape is the right direction for Claude-first use: `generate_start` returns quickly, job state is persisted, and `generate_status` gives a resumable polling API instead of forcing every render through one long blocking request (`mcp-servers/codex-image2/server.py:448-512,515-533,642-705`).
- The docs are honest about prerequisites and instability: requiring a signed-in local Codex app plus `codex debug app-server send-message-v2 "ping"` is the right level of explicitness for an experimental bridge (`mcp-servers/codex-image2/README.md:13-18,51-55`).
- Keeping this as a separate skill/backend option is the right ecosystem fit; it reads as an opt-in alternative rather than an unsafe replacement for the Gemini path (`skills/paper-illustration-image2/SKILL.md:2-5,208-222`).

## 🛑 Changes required before merge

**1. Worker lifecycle / status can stick forever** (`mcp-servers/codex-image2/server.py:491-499,515-533`) — **Pre-merge hard blocker**.
Minimum acceptable fix: replace the current `os.kill(pid, 0)`-only liveness check with child-aware reaping in the parent process. The smallest correct version: keep `workerPid`, and in `generate_status` call a helper that first tries `os.waitpid(pid, os.WNOHANG)`; if it returns the child (or `ChildProcessError` / missing pid), mark the job failed unless a terminal result is already written. I'd also persist a simple deadline (`timeoutSec` / `expiresAt`) in the job JSON so a stale `running` job cannot survive indefinitely after a bad exit or restart.
Rationale: a dead worker should become `failed`, not "maybe alive forever".

**2. Output writes are not sandboxed to the workspace** (`mcp-servers/codex-image2/server.py:178-185,300-311,466-481,725-735`; `skills/paper-illustration-image2/SKILL.md:214-215,254-256`) — **Pre-merge hard blocker**.
Minimum acceptable fix: after `resolve_output_path`, require the resolved path to live under `cwd/figures/ai_generated/` (or, if you want a slightly broader contract, at least under `cwd`). If it escapes that root, return `tool_error` instead of writing. Concrete shape: `allowed_root = (cwd / "figures" / "ai_generated").resolve()` and reject if `resolved.relative_to(allowed_root)` fails.
Rationale: this tool should not be able to write arbitrary files outside the paper workspace.

**3. The long-hang class is still exposed** (`mcp-servers/codex-image2/server.py:31,344-359,642-705,715-738,760-765`) — **Pre-merge hard blocker**.
Minimum acceptable fix: pick one of two paths and make it explicit. Preferred: remove synchronous `generate` entirely and expose only `generate_start` + `generate_status`. If you keep `generate`, then cap it aggressively (for example, pass a short timeout override such as 60s) and hard-cap `waitSeconds` (for example `min(max(int(waitSeconds), 0), 30)`) so one MCP call cannot occupy the server for 10 minutes. Please also update the README/skill to steer callers to async.
Rationale: once Codex stalls, the MCP server must stay responsive. (This is the same failure class as #151 in this repo — we don't want to amplify it.)

**4. The new skill does not yet satisfy the integration contract** (`skills/paper-illustration-image2/SKILL.md:148-156,224-256`; `skills/shared-references/integration-contract.md:34-35,37-60,73-105,142-150`) — **Pre-merge hard blocker**.
Minimum acceptable fix: add one small canonical helper script, e.g. `tools/paper_illustration_image2.py`, with three subcommands: `preflight`, `finalize`, and `verify`. `preflight` checks the observable predicate (`codex` exists and `codex debug app-server send-message-v2 "ping"` succeeds); `finalize` copies the chosen best image to `figure_final.png` and writes `latex_include.tex` + `review_log.json`; `verify` emits a small JSON diagnostic for those artifacts. Then update `SKILL.md` to include a real checkbox block for the long workflow, and add a row for this integration to the contract table.
Rationale: right now the workflow describes the integration, but it does not leave a reliable receipt or repair path if the model skips a step.

**5. Tests cover parser basics, but not the risky failure paths** (`tests/test_codex_image2_server.py:75-115`; `mcp-servers/codex-image2/server.py:328-391,491-533`) — **Pre-merge hard blocker**.
Minimum acceptable fix: add unit tests around `run_codex_image()` / status handling by mocking `subprocess.run` / `Popen`. The minimum set I'd ask for: `codex` missing, `TimeoutExpired`, `NATIVE_IMAGE_UNAVAILABLE`, rejection of `commandExecution`, malformed `imageGeneration` payload, and the dead-worker / stale-job path after the lifecycle fix. Please also stop treating arbitrary base64 bytes as a valid image success case; the current test accepts `b"hello"` as `out.png` (`tests/test_codex_image2_server.py:99-115`).
Rationale: the tricky logic in this PR is all in the failure paths, so that is what needs tests.

## 💡 Additional requests

- Please make debug logging opt-in or redact prompt/system by default (`mcp-servers/codex-image2/server.py:33-35,49-55,361-366,468-485,612`). Right now raw prompts, paths, and stdout/stderr are persisted to `/tmp` and `~/.claude/state/...`; for an OAuth-backed local bridge, that is more data retention than I think we should default to.
- Please tighten success validation so a non-image blob does not count as a PNG success (`mcp-servers/codex-image2/server.py:304-312`; `tests/test_codex_image2_server.py:99-115`). A tiny signature check (PNG magic `\x89PNG\r\n\x1a\n`) would be enough.
- I'd save the broader rename (`image2`), the debug-trace fragility discussion, and richer OAuth preflight/recovery UX for a follow-up PR. They're real concerns, but not the minimum merge gate once the five items above are fixed.

## ❓ Design question for you

Do you actually need the synchronous `generate` tool, or can this bridge be async-only (`generate_start` + `generate_status`)? That's the most load-bearing design choice here: if the sync tool is unnecessary, removing it is the cleanest way to avoid reintroducing the long-hang class from #151.

## 📌 Context

This review was produced via the ARIS cross-model protocol: GPT-5.4 via Codex MCP read the PR files independently without executor framing, per `skills/shared-references/reviewer-independence.md`. I'm the maintainer relaying consolidated feedback — happy to clarify any of the blockers above if the minimum-fix wording is unclear, either here or on a follow-up thread.

Thanks again for the serious work here — looking forward to v2.

kbr19-thu:
Thanks — I pushed a follow-up hardening pass that addresses the pre-merge blockers in `d334505`.

What changed:

- Worker lifecycle: jobs now persist `timeoutSec` + `expiresAt`, and `generate_status` does child-aware worker probing plus deadline failure instead of relying only on `os.kill(pid, 0)`.
- Output sandbox: writes are now restricted to `cwd/figures/ai_generated/`; escaped `outputPath` values are rejected.
- Long-hang class: I removed the synchronous `generate` tool entirely. The bridge is now async-only (`generate_start` + `generate_status`), and `waitSeconds` is capped to keep the MCP server responsive.
- Integration contract: added `tools/paper_illustration_image2.py` with `preflight`, `finalize`, and `verify`, updated the skill with a visible checklist + repair path, and added the integration row to `skills/shared-references/integration-contract.md`.
- Tests: expanded unit coverage for codex missing, timeout, native-unavailable, commandExecution rejection, malformed/non-PNG payloads, output sandboxing, and stale/dead worker failure paths.

Additional hardening included:

- debug logging is now opt-in via `CODEX_IMAGE2_DEBUG_LOG`
- raw run logs are off by default unless `CODEX_IMAGE2_SAVE_RUN_LOGS=1`
- PNG signature validation is required before a generated blob is accepted as success

Validation rerun on this branch:

- `python3 -m unittest tests/test_codex_image2_server.py`
- `python3 -m py_compile mcp-servers/codex-image2/server.py tools/paper_illustration_image2.py`
- `git diff --check`
- helper smoke: `finalize` + `verify`
- live native bridge smoke after the fixes (`generate_start` + `generate_status`)

And to answer the design question directly: no, I do not think the synchronous `generate` tool is worth keeping here. This revision removes it and keeps the bridge async-only.

wanshuiyin (approved):
v2 LGTM. Went through all 5 blockers line-by-line — every minimum-fix shape matches exactly what we asked for, and you proactively handled the additional requests (debug-log opt-in via `CODEX_IMAGE2_DEBUG_LOG`, PNG magic validation, removing sync `generate` as a clean design call).

Specific verifications I re-ran:
- `get_generate_status` double-gate: `expiresAt` deadline + `classify_worker_state` (with `os.waitpid(pid, os.WNOHANG)`) — dead workers do become `failed`
- `validate_output_path` wired in at both call sites (lines ~433 and ~578), rejects anything outside `cwd/figures/ai_generated/` via `relative_to` catch
- `"name": "generate"` entirely gone from the tool registry; only `generate_start` + `generate_status` remain
- `tools/paper_illustration_image2.py` has real implementations of all three subcommands — `preflight` actually shells out to `codex debug app-server send-message-v2 ping`, `finalize` writes all three artifacts (`figure_final.png` + `latex_include.tex` + `review_log.json`), `verify` checks PNG magic and cross-references
- Tests: the v1 "b'hello' becomes a valid PNG" embarrassment is replaced with `test_materialize_generated_image_rejects_non_png_base64` — that proactive correction was appreciated

Cleanest external PR revision I've reviewed in a while. Approving.

PR #171, opened 2026-04-25
Support Claude Code object-style hooks

Jxy-yxJ:
This change lets ARIS load both legacy string-array hooks and newer Claude Code object-style hooks. Object-style command hooks are flattened into the existing RuntimeHookConfig command list, preserving existing hook runner behavior while avoiding settings load failures such as `field PostToolUse must contain only strings`.

Also gates Unix-only test chmod logic so runtime tests compile on Windows.

### DIFF
--- a/crates/runtime/src/config.rs
+++ b/crates/runtime/src/config.rs
@@ -496,9 +496,9 @@ fn parse_optional_hooks_config(root: &JsonValue) -> Result<RuntimeHookConfig, Co
     };
     let hooks = expect_object(hooks_value, "merged settings.hooks")?;
     Ok(RuntimeHookConfig {
-        pre_tool_use: optional_string_array(hooks, "PreToolUse", "merged settings.hooks")?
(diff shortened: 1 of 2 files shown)


wanshuiyin:
Hi @Jxy-yxJ — your object-style hooks parser is shipping in **v0.4.5** 🎉 (squashed as `7edcafc`).

This fixes a quietly-bad failure mode: a Claude Code-style hooks config like

```json
"hooks": {
  "PreToolUse": [
    { "matcher": ".*", "hooks": [{ "type": "command", "command": "..." }] }
  ]
}
```

used to make the entire `feature_config` fall back to default, silently wiping any user-configured MCP servers / OAuth / sandbox / permissionMode. Object-style hooks now parse natively, and the rest of `HookRunner` is unchanged because we flatten them to the existing command-list shape.

Verified: `cargo test -p runtime` 28/28 pass (incl. your new `loads_object_style_claude_code_hooks_as_commands`).

Thanks for the clean, focused PR.

— maintainer team

PR #212, opened 2026-05-08
fix(install): auto-detect Codex CLI platform and delegate to install_aris_codex.sh (#180)

### LINKED ISSUE #180: install_aris.sh无法自动适配Codex CLI
现在使用aris_repo/tools/install_aris.sh默认安装适配CLAUDE，似乎在最近一次commit中去掉了平台适配

dashitongzhi:
## Summary

Restores platform auto-detection for Codex CLI users in `install_aris.sh` (#180).

When the target project has Codex markers (`.agents/`, `AGENTS.md`, `.codex/config.toml`) and no Claude markers (`.claude/`, `CLAUDE.md`), `install_aris.sh` now automatically delegates to `install_aris_codex.sh`.

## Changes

- **`auto_detect_platform()`**: New function that inspects the project directory for platform markers
- **`--platform` option restored**: Accepts `codex` or `claude` for explicit override (previously errored with "removed")
- **Arg filtering**: `--platform` is stripped from args before delegation to avoid unknown-option errors in `install_aris_codex.sh`
- **Validation**: Invalid `--platform` values are rejected with a clear error

## Behavior

| Project markers | Behavior |
|---|---|
| Codex only (`.agents/`, `AGENTS.md`) | Auto-delegates to `install_aris_codex.sh` |
| Claude only (`.claude/`, `CLAUDE.md`) | Runs normal Claude install |
| Both or neither | Runs normal Claude install (no auto-detect) |
| `--platform codex` | Forces Codex delegation regardless of markers |
| `--platform claude` | Forces Claude install regardless of markers |

## Testing

```
# Codex project: auto-detected and delegated
$ bash tools/install_aris.sh /tmp/test-codex-project --dry-run
Codex CLI platform detected — delegating to install_aris_codex.sh
...

# Claude project: normal flow
$ bash tools/install_aris.sh /tmp/test-claude-project --dry-run
ARIS Project Install
...

# Override: force codex on Claude project
$ bash tools/install_aris.sh /tmp/test-claude-project --platform codex --dry-run
Codex CLI platform detected — delegating to install_aris_codex.sh
...

# Invalid value rejected
$ bash tools/install_aris.sh /tmp/test --platform invalid
Error: --platform must be codex or claude (got: invalid)
```

## Closes

Fixes #180

### DIFF
--- a/tools/install_aris.sh
+++ b/tools/install_aris.sh
@@ -24,6 +24,7 @@
 #                          the correct upstream target (repeatable)
 #   --replace-link NAME    replace an upstream-internal symlink that points to
 #                          a DIFFERENT entry than expected (repeatable)
+#   --platform PLATFORM    force platform: codex or claude (default: auto-detect)
 #   --from-old             trigger migration from legacy nested install
 #                          (.claude/skills/aris/)
 #   --migrate-copy STRAT   for legacy COPY install: STRAT = keep-user | prefer-upstream
@@ -82,6 +83,31 @@ REPLACE_LINK_NAMES=()
 
 usage() { sed -n '2,40p' "$0" | sed 's/^# \?//'; }
 
+ORIGINAL_ARGS=("$@")
+PLATFORM_OVERRIDE=""
+DETECTED_PLATFORM=""
+
+# ─── Platform auto-detection for Codex CLI (#180) ──────────────────────────────
+# If the project has Codex markers (.agents/, AGENTS.md, .codex/config.toml)
+# and NO Claude markers (.claude/, CLAUDE.md), delegate to install_aris_codex.sh.
+auto_detect_platform() {
+    local proj="$1"
+    local has_codex_markers=false
+    local has_claude_markers=false
+
+    # Codex markers
+    [[ -d "$proj/.agents" || -f "$proj/AGENTS.md" || -f "$proj/.codex/config.toml" ]] && has_codex_markers=true
+
+    # Claude markers
+    [[ -d "$proj/.claude" || -f "$proj/CLAUDE.md" || -f "$proj/.claude/settings.json" ]] && has_claude_markers=true
+
+    if $has_codex_markers && ! $has_claude_markers; then
+        DETECTED_PLATFORM="codex"
+    elif $has_claude_markers; then
+        DETECTED_PLATFORM="claude"
+    fi
+}
+
 while [[ $# -gt 0 ]]; do
     case "$1" in
         --reconcile)         ACTION="reconcile"; shift ;;
@@ -96,9 +122,7 @@ while [[ $# -gt 0 ]]; do
         --adopt-existing)    ADOPT_NAMES+=("${2:?--adopt-existing requires NAME}"); shift 2 ;;
         --replace-link)      REPLACE_LINK_NAMES+=("${2:?--replace-link requires NAME}"); shift 2 ;;
         --platform)
-            echo "Error: --platform is removed. ARIS now only supports Claude Code (.claude/skills/)." >&2
-            echo "       Codex CLI users: see docs for the manual codex setup." >&2
-            exit 2 ;;
+            PLATFORM_OVERRIDE="${2:?--platform requires codex|claude}"; shift 2 ;;
         --force)
             echo "Error: --force is removed. Use the granular flags:" >&2
             echo "       --adopt-existing NAME (for non-managed symlinks pointing to correct upstream)" >&2
@@ -116,6 +140,9 @@ done
 
 if [[ -n "$MIGRATE_COPY" && "$MIGRATE_COPY" != "keep-user" && "$MIGRATE_COPY" != "prefer-upstream" ]]; then
     echo "Error: --migrate-copy must be keep-user or prefer-upstream (got: $MIGRATE_COPY)" >&2; exit 2
+fi
+if [[ -n "$PLATFORM_OVERRIDE" && "$PLATFORM_OVERRIDE" != "codex" && "$PLATFORM_OVERRIDE" != "claude" ]]; then
+    echo "Error: --platform must be codex or claude (got: $PLATFORM_OVERRIDE)" >&2; exit 2
 fi
 
 # ─── Helpers ──────────────────────────────────────────────────────────────────
@@ -249,6 +276,27 @@ MANIFEST_PATH="$PROJECT_ARIS_DIR/$MANIFEST_NAME"
 MANIFEST_PREV="$PROJECT_ARIS_DIR/$MANIFEST_PREV_NAME"
 LOCK_DIR="$PROJECT_ARIS_DIR/$LOCK_DIR_NAME"
 DOC_FILE="$PROJECT_PATH/$DOC_FILE_NAME"
+
+auto_detect_platform "$PROJECT_PATH"
+PLATFORM="${PLATFORM_OVERRIDE:-$DETECTED_PLATFORM}"
+if [[ "$PLATFORM" == "codex" ]]; then
+    SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+    CODEX_INSTALLER="$SCRIPT_DIR/install_aris_codex.sh"
+    [[ -f "$CODEX_INSTALLER" ]] || die "Codex installer not found: $CODEX_INSTALLER"
+    log ""
+    log "Codex CLI platform detected — delegating to install_aris_codex.sh"
+    log "  (override with --platform claude)"
+    log ""
+    # Strip --platform flag from args before delegating
+    FILTERED_ARGS=()
+    skip_next=false
+    for arg in "${ORIGINAL_ARGS[@]}"; do
+        if $skip_next; then skip_next=false; continue; fi
+        if [[ "$arg" == "--platform" ]]; then skip_next=true; continue; fi
+        FILTERED_ARGS+=("$arg")
+    done
+    exec bash "$CODEX_INSTALLER" "${FILTERED_ARGS[@]}"
+fi
 
 # ─── S9: refuse if .aris / .claude / .claude/skills is itself a symlink ───────
 # (.aris and .claude/skills may not exist yet — only check if present.)
--- a/tools/install_aris.sh
+++ b/tools/install_aris.sh
@@ -86,6 +86,8 @@ usage() { sed -n '2,40p' "$0" | sed 's/^# \?//'; }
 ORIGINAL_ARGS=("$@")
 PLATFORM_OVERRIDE=""
 DETECTED_PLATFORM=""
+CLAUDE_ONLY_FLAGS_USED=()
+CODEX_ONLY_FLAGS_SEEN=()
 
 # ─── Platform auto-detection for Codex CLI (#180) ──────────────────────────────
 # If the project has Codex markers (.agents/, AGENTS.md, .codex/config.toml)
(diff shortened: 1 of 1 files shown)


wanshuiyin (changes requested):
Thanks for restoring `--platform` and adding auto-detect — this directly addresses #180. The detect heuristic and `exec` delegation look right; one remaining routing edge keeps `--platform codex` from being fully usable, though.

`install_aris.sh` parses the full argv before checking platform / delegating, which creates two asymmetric cases:

**Case 1 (bug) — Codex-only flags hard-fail at the parent**

```bash
bash tools/install_aris.sh /path --platform codex --with-gemini-review-overlay
# Unknown option: --with-gemini-review-overlay
```

`install_aris_codex.sh` accepts `--with-claude-review-overlay` and `--with-gemini-review-overlay`, but the parent doesn't know them and rejects them before delegating. So `--platform codex` is effectively unusable together with overlay flags.

**Case 2 (UX) — Claude-only flags get silently forwarded then fail in the child**

```bash
bash tools/install_aris.sh /path --platform codex --from-old
# install_aris_codex.sh: Unknown option: --from-old
```

The parent accepts `--from-old`, then `exec`s the codex installer with the original args, which doesn't recognize `--from-old`. User gets an unclear error from the child script.

### Suggested fix: classify flags in the parent and route accordingly

```bash
ORIG_ARGS=("$@")
SAW_CODEX_ONLY=false
CLAUDE_ONLY_USED=()

while [[ $# -gt 0 ]]; do
  case "$1" in
    --platform)
      PLATFORM_OVERRIDE="${2:?--platform requires codex|claude}"; shift 2 ;;

    # common (both installers accept)
    --reconcile|--uninstall|--dry-run|--quiet|--no-doc|--clear-stale-lock)
      # ... existing handling
      shift ;;
    --aris-repo|--replace-link)
      # ... existing handling
      shift 2 ;;

    # Codex-only — recognize & consume so parent doesn't reject, but don't act
    --with-claude-review-overlay|--with-gemini-review-overlay)
      SAW_CODEX_ONLY=true
      shift ;;

    # Claude-only — track to error in codex mode
    --from-old)
      CLAUDE_ONLY_USED+=("$1"); FROM_OLD=true; shift ;;
    --migrate-copy|--adopt-existing)
      CLAUDE_ONLY_USED+=("$1"); # ... existing parsing
      shift 2 ;;

    --force) # ... existing ;;
    --*) echo "Unknown option: $1" >&2; exit 2 ;;
    *)   # project path handling as today
         shift ;;
  esac
done

# ... auto_detect_platform after parsing

if [[ "$PLATFORM" == "codex" ]]; then
  if ((${#CLAUDE_ONLY_USED[@]})); then
    echo "Error: ${CLAUDE_ONLY_USED[*]} are Claude-only options and cannot be used with --platform codex." >&2
    echo "       Codex installer supports: --with-claude-review-overlay, --with-gemini-review-overlay." >&2
    exit 2
  fi

  FILTERED_ARGS=()
  skip_next=false
  for arg in "${ORIG_ARGS[@]}"; do
    if \$skip_next; then skip_next=false; continue; fi
    if [[ "\$arg" == "--platform" ]]; then skip_next=true; continue; fi
    FILTERED_ARGS+=("\$arg")
  done
  exec bash "\$SCRIPT_DIR/install_aris_codex.sh" "\${FILTERED_ARGS[@]}"
fi

if [[ "\$PLATFORM" == "claude" && "\$SAW_CODEX_ONLY" == true ]]; then
  echo "Error: --with-{claude,gemini}-review-overlay are Codex-only; use --platform codex." >&2
  exit 2
fi
```

### Dual-marker warning

For the auto-detect path when both `.claude/` and `.codex/` markers exist, a heads-up before the default-Claude fallback would help users with mixed setups:

```
warning: both .claude/ and .codex/ markers found; defaulting to Claude installer for backward compatibility.
         Pass --platform codex to install ARIS for Codex CLI instead.
```

### Flag inventory for reference

- **common (both):** `--reconcile` / `--uninstall` / `--aris-repo` / `--dry-run` / `--quiet` / `--no-doc` / `--replace-link` / `--clear-stale-lock` (8)
- **Claude-only:** `--from-old` / `--migrate-copy` / `--adopt-existing` (3)
- **Codex-only:** `--with-claude-review-overlay` / `--with-gemini-review-overlay` (2)

Once arg routing is symmetric I'd be happy to merge — thanks again for picking this up.

dashitongzhi:
## Fix for review feedback

I've addressed all 4 review issues:

1. **Codex-only flags** (--with-claude-review-overlay, --with-gemini-review-overlay) are now recognized by parent and not rejected
2. **Claude-only flags** (--from-old, --migrate-copy, --adopt-existing) now error when used with --platform codex
3. **Auto-detect + delegation** moved before resolve_aris_repo
4. **Log message** now says 'selected' for explicit --platform, 'detected' for auto-detect
5. **Dual-marker warning** added when both .claude/ and .codex/ exist

Patch attached below.

wanshuiyin (changes requested):
Thanks @dashitongzhi — substantive iteration. The arg-routing concerns from my prior review are mostly addressed (claude-only flag tracking, codex-only flag tracking, `--platform` validation, `exec bash` delegation pattern). But cross-model review surfaced **3 blocking issues** that need fixing before merge:

## Blocker 1: Fresh Codex project silently routes to Claude

`auto_detect_platform` only sets `DETECTED_PLATFORM="codex"` when `.agents/`, `AGENTS.md`, or `.codex/config.toml` exists. A **fresh Codex CLI project** (just initialized, no `.agents/` yet, no `AGENTS.md` written) lacks all three markers. With no markers and no `--platform`, `DETECTED_PLATFORM` stays empty, `PLATFORM` is empty, the codex delegation branch is skipped, and the script proceeds to install at `.claude/skills/`. That's wrong — silent install to the wrong path.

Suggested fix: when no markers are detected and `--platform` is not passed, exit with an explicit message asking for `--platform claude|codex`. Don't default-silently-Claude.

## Blocker 2: Both-markers ambiguity defaults to Claude

```bash
if $has_codex_markers && $has_claude_markers; then
    warn "Both Claude and Codex markers found in $proj"
    warn "  Defaulting to claude; use --platform codex to override."
    DETECTED_PLATFORM="claude"
```

A user with both `CLAUDE.md` and `AGENTS.md` who runs `install_aris.sh` expecting to install Codex skills gets Claude. The warn-then-proceed pattern is risky for an installer that touches a project directory. Suggested fix: when both markers present and no `--platform` passed, **exit** with the same require-explicit message — don't proceed.

## Blocker 3 (critical): \`--quiet\` semantics diverge between installers

Verified by reading both files:

- `tools/install_aris.sh` `prompt()`: `\$QUIET && return 1` — in quiet mode, prompts return **false** (ABORT-on-prompt)
- `tools/install_aris_codex.sh` `prompt()`: `\$QUIET && return 0` — in quiet mode, prompts return **true** (AUTO-ACCEPT-on-prompt)

The parent forwards `--quiet` to the codex installer via the `FILTERED_ARGS` exec. A user running `install_aris.sh --quiet` on a project that auto-detects as codex gets silent auto-accept of every prompt, instead of the abort-on-prompt semantics they'd get under Claude routing. That's a real safety divergence.

Fix options:
1. Unify `prompt()` semantics in both installers (preferred — pick one convention)
2. Parent layer rejects `--quiet` when delegating to codex installer with a message asking the user to invoke `install_aris_codex.sh` directly
3. Parent layer maps `--quiet` to the appropriate flag for each installer (more invasive)

## Passes (verified working)

- Claude-only flag tracking (`--from-old`, `--migrate-copy`, `--adopt-existing`) is correct: these all touch legacy `.claude/skills/aris/` migration paths.
- `--replace-link` and `--clear-stale-lock` correctly NOT in `CLAUDE_ONLY_FLAGS_USED` — both installers parse them.
- `--with-claude-review-overlay` and `--with-gemini-review-overlay` are actually consumed downstream in `install_aris_codex.sh` (overlay package selection in `selected_packages()`).
- `--platform` order check happens after parsing, before delegation — good order.
- `exec bash` delegation pattern is clean.

## Minor (non-blocking)

- Multiple `--platform` flags: current behavior is last-wins. Consider rejecting duplicates explicitly.
- `--platform=value` form falls through to `Unknown option`. Acceptable but worth a `--help` note.
- Duplicate entries in `CLAUDE_ONLY_FLAGS_USED` (e.g., user passes `--adopt-existing X --adopt-existing Y` → message says `... --adopt-existing --adopt-existing`). Cosmetic.

## Suggestion

The fresh-Codex-project fallback issue (Blocker 1) and the `--quiet` divergence (Blocker 3) are the substantive ones. Once those are fixed, this is mergeable. The both-markers exit (Blocker 2) is a smaller change but in the same area.

Happy to merge after these three are addressed.

wanshuiyin (changes requested):
Thanks — auto-detecting the platform and delegating to `install_aris_codex.sh` is a nice UX win. Two blockers before merge though (both independently confirmed by a Claude + GPT-5.5 xhigh cross-model review), plus one design check.

## 🔴 P1 — bash 3.2 + `set -u`: empty-array expansion aborts the delegation
The script runs `set -euo pipefail` under `#!/usr/bin/env bash`, and macOS ships /bin/bash 3.2.57. The new block does `for arg in "${ORIGINAL_ARGS[@]}"` and ends with `exec bash "$CODEX_INSTALLER" "${FILTERED_ARGS[@]}"`. Under bash 3.2 + `set -u`, expanding an **empty** array throws `unbound variable` and aborts — so running the installer with no extra args from a Codex-marked project dies *before* it can delegate (the exact feature this PR adds). The file already uses the safe pattern elsewhere (`[[ ${#arr[@]} -gt 0 ]]` guard before iterating).
→ Guard both expansions, e.g. `"${ORIGINAL_ARGS[@]+"${ORIGINAL_ARGS[@]}"}"`, or gate the loop/exec on a non-empty count.

## 🔴 P1 — detection isn't fail-closed: a Claude user can be silently mis-routed
`AGENTS.md` alone is treated as a Codex marker. But a fresh Claude project usually has no `.claude/` yet (the installer creates it), so a repo carrying `AGENTS.md` but no `CLAUDE.md`/`.claude/` yields has_codex=true, has_claude=false and is silently `exec`'d into the Codex installer — wrong skill set, no confirmation prompt, only a `log`-level message (suppressed under `--quiet`).
→ Require explicit `--platform` (or a confirmation) when no Claude marker exists, or only auto-delegate on stronger Codex markers (`.agents/` / `.codex/config.toml`), not `AGENTS.md` alone.

## 🟡 P2 — `--platform` stripping is token-based
A forwarded argument whose *value* is literally `--platform` would be removed along with the following token. Low likelihood, but avoidable by rebuilding the forwarded args during the real option parse.

## Design check
`--platform` was deliberately removed earlier (with an explicit error pointing Codex users to manual setup), so re-adding it reverses that decision. That's a legitimate choice — just please confirm it's intended, and double-check the inline usage/help text still renders coherently with the re-added flag.

The diff applies cleanly to current `main` and needs no inventory integration (no skills added). Fix the two P1s and this is good to go. Thanks for the contribution!

wanshuiyin (approved):
Approving — the three blockers from the prior round are all fixed, verified by a Claude + GPT-5.5 xhigh cross-model pass:

- **P1a** (bash 3.2 + `set -u` empty-array abort): the `exec` is now guarded by `[[ ${#FORWARDED_ARGS[@]} -gt 0 ]]` with a separate empty-arg `exec`, and the two `[*]` die-messages sit inside non-empty count guards. Confirmed against `/bin/bash 3.2.57`.
- **P1b** (fail-closed detection): `AGENTS.md` alone is a *hint* only now — it warns and stays on `claude`; only `.agents/` / `.codex/config.toml` auto-delegate to Codex.
- **P2** (`--platform` token-stripping): `--platform` is consumed during the real parse and never forwarded, so a forwarded value literally `--platform` can't be mis-stripped.

Integration checked against `install_aris_codex.sh`: every flag forwarded on the codex path is accepted there, and the claude-only flags (`--from-old` / `--migrate-copy` / `--adopt-existing`) are blocked before delegation.

One cosmetic non-blocker (fine as a tiny follow-up): with an explicit `--platform codex` in a dual-marker / `AGENTS.md`-only project, `auto_detect_platform` still prints a "defaulting to claude" warning before the override is applied — behaviour is correct, only the message is misleading.

Thanks for the careful iteration across rounds. Merging — closes #180.

PR #220, opened 2026-05-13
fix(claude-review): parse JSON-array output from claude CLI 2.x

treerobin06:
## Summary

`claude -p --output-format json` switched from NDJSON-of-dicts to a single JSON array on claude CLI 2.x. `parse_claude_json` in `mcp-servers/claude-review/server.py` only handled the dict case, so every `review` / `review_reply` call on a modern claude CLI returns `Claude CLI did not return JSON output` even though the CLI succeeded.

## Repro

```bash
# Pre-patch behavior on claude CLI 2.1.140
$ python3 mcp-servers/claude-review/server.py < initialize-and-call.jsonl
# tools/call result -> {"error": "Claude CLI did not return JSON output"}, isError: true

$ claude -p OK --output-format json --permission-mode plan | head -c 80
[{"type":"system","subtype":"init",...
# Single-line JSON array, not NDJSON.
```

The 4 array members are `system/init`, `assistant`, `rate_limit_event`, `result` — `result` is the dict the old code path was implicitly grabbing as the last NDJSON line.

## Fix

When `json.loads(line)` returns a `list`, walk it in reverse:
1. Prefer the terminal `type == "result"` event (semantically what `run_claude_review` consumes downstream — `session_id`, `result`, `model`, `duration_ms`, `stop_reason`).
2. Fall back to the last dict in the array, in case the event shape changes again.

NDJSON output still parses via the unchanged `dict` branch, so older claude CLI versions keep working.

## Verification

Tested locally with claude CLI 2.1.140:

```
review tools/call → {
  "threadId": "23d9b3ae-4bd3-4be2-82a6-1820fd035bdc",
  "response": "OK",
  "duration_ms": 16656,
  "stop_reason": "end_turn"
}
isError: false
```

## Out of scope

`mcp-servers/gemini-review/server.py` has a structurally identical `parse_gemini_json`. I didn't touch it because I haven't reproduced the same shape with current gemini CLI — flagging here so a maintainer with gemini-review locally can confirm and apply the same shape if needed.

## PR Checklist

- [x] Code follows the project style (minimal, focused, comment explains the rationale)
- [x] No public API change — same `(payload_dict, error)` return contract
- [x] Tested locally against claude CLI 2.1.140 — `review` and `review_reply` end-to-end
- [ ] Documentation updated — not applicable, internal parser change

### DIFF
--- a/mcp-servers/claude-review/server.py
+++ b/mcp-servers/claude-review/server.py
@@ -142,7 +142,11 @@ def parse_claude_json(raw_stdout: str) -> tuple[dict[str, Any] | None, str | Non
                 return item, None
         return None, "Claude CLI returned a JSON array without a 'result' event"
 
-    # Legacy CLI 1.x: NDJSON stream of dicts, walk lines in reverse for the last dict.
+    # Legacy CLI 1.x: NDJSON stream of dicts, walk lines in reverse for the
+    # last useful payload. Same array-vs-dict policy as above so a CLI 2.x
+    # JSON-array line surrounded by non-JSON noise (wrapper warnings, nvm/asdf
+    # banners, future CLI debug prints) still surfaces the result event
+    # instead of being silently dropped.
     for candidate in reversed(stripped.splitlines()):
         candidate = candidate.strip()
         if not candidate:
@@ -153,6 +157,12 @@ def parse_claude_json(raw_stdout: str) -> tuple[dict[str, Any] | None, str | Non
             continue
         if isinstance(line_payload, dict):
             return line_payload, None
+        if isinstance(line_payload, list):
+            for item in reversed(line_payload):
+                if isinstance(item, dict) and item.get("type") == "result":
+                    return item, None
+            # fall through: this line was an array without a result event,
+            # but earlier lines might still carry one — keep scanning.
 
     return None, "Claude CLI did not return JSON output"
 
--- a/mcp-servers/claude-review/server.py
+++ b/mcp-servers/claude-review/server.py
@@ -108,28 +108,42 @@ def find_claude_bin() -> str | None:
 
 
 def parse_claude_json(raw_stdout: str) -> tuple[dict[str, Any] | None, str | None]:
-    lines = [line.strip() for line in raw_stdout.splitlines() if line.strip()]
-    if not lines:
+    stripped = raw_stdout.strip()
+    if not stripped:
         return None, "Claude CLI returned empty output"
 
-    for candidate in reversed(lines):
+    # CLI 2.x: try whole stdout as a single JSON value first.
+    # Handles both compact one-line arrays and pretty-printed multi-line arrays.
+    try:
+        payload = json.loads(stripped)
+    except json.JSONDecodeError:
+        payload = None
+
+    if isinstance(payload, dict):
+        return payload, None
+    if isinstance(payload, list):
+        # claude CLI 2.x emits a JSON array of event objects under --output-format json
+        # (system init -> assistant -> rate_limit_event -> result). Only the terminal
+        # "result" event carries the fields run_claude_review consumes downstream
+        # (result text, session_id, duration_ms, stop_reason). Falling back to any
+        # other dict (e.g. rate_limit_event) would surface as a "successful parse"
+        # producing an empty review — strictly worse than a clear error.
+        for item in reversed(payload):
+            if isinstance(item, dict) and item.get("type") == "result":
+                return item, None
+        return None, "Claude CLI returned a JSON array without a 'result' event"
+
+    # Legacy CLI 1.x: NDJSON stream of dicts, walk lines in reverse for the last dict.
+    for candidate in reversed(stripped.splitlines()):
+        candidate = candidate.strip()
+        if not candidate:
+            continue
         try:
-            payload = json.loads(candidate)
+            line_payload = json.loads(candidate)
         except json.JSONDecodeError:
             continue
-        if isinstance(payload, dict):
-            return payload, None
-        if isinstance(payload, list):
-            # claude CLI 2.x emits a single JSON array of event objects under
-            # --output-format json (system init -> assistant -> rate_limit -> result),
-            # not the NDJSON-style dict stream this parser originally expected.
-            # Prefer the terminal "result" event; fall back to the last dict.
-            for item in reversed(payload):
-                if isinstance(item, dict) and item.get("type") == "result":
-                    return item, None
-            for item in reversed(payload):
-                if isinstance(item, dict):
-                    return item, None
+        if isinstance(line_payload, dict):
+            return line_payload, None
 
     return None, "Claude CLI did not return JSON output"
 
--- a/mcp-servers/claude-review/server.py
+++ b/mcp-servers/claude-review/server.py
@@ -119,6 +119,17 @@ def parse_claude_json(raw_stdout: str) -> tuple[dict[str, Any] | None, str | Non
             continue
         if isinstance(payload, dict):
             return payload, None
+        if isinstance(payload, list):
+            # claude CLI 2.x emits a single JSON array of event objects under
+            # --output-format json (system init -> assistant -> rate_limit -> result),
+            # not the NDJSON-style dict stream this parser originally expected.
+            # Prefer the terminal "result" event; fall back to the last dict.
+            for item in reversed(payload):
+                if isinstance(item, dict) and item.get("type") == "result":
+                    return item, None
+            for item in reversed(payload):
+                if isinstance(item, dict):
+                    return item, None
 
     return None, "Claude CLI did not return JSON output"
 
--- a/mcp-servers/claude-review/server.py
+++ b/mcp-servers/claude-review/server.py
@@ -20,8 +20,17 @@ from pathlib import Path
 from typing import Any
 
 
-sys.stdout = os.fdopen(sys.stdout.fileno(), "wb", buffering=0)
-sys.stdin = os.fdopen(sys.stdin.fileno(), "rb", buffering=0)
+def _configure_stdio_for_mcp() -> None:
+    """Re-bind sys.stdout/sys.stdin as raw binary streams for the MCP protocol.
+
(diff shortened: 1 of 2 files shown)


wanshuiyin (changes requested):
Thanks @treerobin06 — clean problem analysis with a reproducible failure and a backwards-compatible patch. Did a fresh cross-model review pass; two issues to address before merge, plus a small testing ask.

## 1. The patch only handles single-line JSON arrays

`parse_claude_json` splits stdout into lines before parsing:

```python
lines = [line.strip() for line in raw_stdout.splitlines() if line.strip()]
...
for candidate in reversed(lines):
    payload = json.loads(candidate)
```

Your repro shows claude CLI 2.x emitting the array on one line, which works under this loop. But if a future CLI version (or different flag combination) pretty-prints across newlines, each `splitlines()` row becomes a JSON fragment, every `json.loads` raises `JSONDecodeError`, and the parser falls through to the original \"did not return JSON output\" error — silent regression.

Suggested shape (try whole-stdout first, fall back to per-line NDJSON):

```python
def parse_claude_json(raw_stdout: str) -> tuple[dict[str, Any] | None, str | None]:
    stripped = raw_stdout.strip()
    if not stripped:
        return None, \"Claude CLI returned empty output\"

    # CLI 2.x: try whole-stdout JSON first (handles both compact and pretty-printed arrays)
    try:
        payload = json.loads(stripped)
    except json.JSONDecodeError:
        payload = None

    if isinstance(payload, dict):
        return payload, None
    if isinstance(payload, list):
        for item in reversed(payload):
            if isinstance(item, dict) and item.get(\"type\") == \"result\":
                return item, None
        return None, \"Claude CLI returned a JSON array without a 'result' event\"

    # Legacy NDJSON (CLI 1.x): walk lines in reverse for the last dict
    for candidate in reversed(stripped.splitlines()):
        candidate = candidate.strip()
        if not candidate:
            continue
        try:
            payload = json.loads(candidate)
        except json.JSONDecodeError:
            continue
        if isinstance(payload, dict):
            return payload, None

    return None, \"Claude CLI did not return JSON output\"
```

## 2. Don't fall back to \"last dict\" when there's no result event

The second `for item in reversed(payload)` block falls back to any trailing dict if no `type==\"result\"` is present. But `run_claude_review` downstream reads `result`, `session_id`, `duration_ms`, `stop_reason` — fields only the `result` event guarantees. Falling back to e.g. a `rate_limit_event` dict will surface to the caller as a \"successful parse\" producing an empty review, which is strictly worse than returning a clear error. Replace the second loop with an explicit parse error (see suggested shape above).

## 3. Test coverage (nice-to-have)

Adding a small `tests/test_parse_claude_json.py` would make this robust to the next CLI shape change. Cases worth covering:

- Single-line JSON array with terminal `result` event (CLI 2.x default, per your repro)
- Multi-line pretty-printed JSON array (defensive)
- Legacy NDJSON of dicts (CLI 1.x — backward compat)
- Empty stdout
- JSON array with no `result` event → parse error
- JSON array where the only dict is `system/init` → parse error (don't silently return)

## On gemini-review

Agree with your call not to touch `mcp-servers/gemini-review/server.py` here — its CLI branch is a fallback (default is the Gemini API), and we don't have an in-house repro. Keeping this PR scoped.

## Doc reference

[Claude SDK headless docs](https://docs.anthropic.com/en/docs/claude-code/sdk/sdk-headless) describe `--output-format json` as returning a single result object and `stream-json` as the events stream. The single JSON array you're observing isn't covered by either description verbatim — flagging in case it's worth filing upstream after this lands.

Happy to merge once **1 and 2** are addressed; test cases are nice-to-have but not blocking if you'd rather ship the parser fix first.

treerobin06:
Thanks for the careful review @wanshuiyin — both points absolutely right, the silent-regression risk on pretty-printed output and the `result`-field guarantee both convinced me. Pushed `8fc5189` addressing all three:

**1. Whole-stdout-first, then NDJSON fallback** — now tries `json.loads(stripped)` on full stdout before line-splitting, so multi-line pretty-printed arrays parse correctly. NDJSON path preserved unchanged for CLI 1.x backwards compat.

**2. Fail-fast on missing result event** — second loop removed. When the array contains no `type=="result"` entry, parser now returns:

```
"Claude CLI returned a JSON array without a 'result' event"
```

instead of silently grabbing a `rate_limit_event` / `system` dict that would surface downstream as an "empty review". Agreed strictly better than silent corruption.

**3. Test coverage** — added `tests/test_claude_review_parse.py` with all six cases you enumerated:

- CLI 2.x single-line array (the original repro)
- Pretty-printed multi-line array (defensive)
- Legacy NDJSON of dicts (CLI 1.x backward compat)
- Empty / whitespace stdout
- Array missing result event → explicit error
- Array containing only `system/init` → explicit error
- Plus non-JSON garbage stdout → falls through to the original error

All 7 pass:

```
$ python3 -m unittest tests.test_claude_review_parse -v
...
Ran 7 tests in 0.011s
OK
```

Re: upstream filing — agreed worth flagging once this lands. The single JSON array shape doesn't match either the `json` (single object) or `stream-json` (NDJSON events) description in the SDK headless docs. Happy to file an Anthropic SDK / claude CLI issue with the verbatim repro after merge if you'd like.

Re: gemini-review — kept out of scope as you suggested. If/when there's an in-house repro for that branch I'm happy to do a follow-up PR with the same parser shape.

treerobin06:
Internal review pass after pushing the v2 patch turned up two follow-up items, addressed in `5566737`:

**MAJOR — fixed:** Module-level `sys.stdout = os.fdopen(..., "wb")` was a global side effect at import time. Reproducible:

```python
spec.loader.exec_module(server_module)
print("hello")  # TypeError: a bytes-like object is required, not 'str'
```

The new tests directly `exec_module(server.py)` to access `parse_claude_json` (matching the existing `test_codex_image2_server.py` pattern), so this would have planted a latent failure for any future test ordering. Moved the stdio rebind into `_configure_stdio_for_mcp()` called from `main()`. `python3 server.py` runtime behavior unchanged; `import server` now non-destructive.

**SUGGESTION — done:** Added 4 mock-`subprocess.run` end-to-end tests for `run_claude_review`, locking the `(threadId, response, model, duration_ms, stop_reason)` field mapping for both CLI 2.x and legacy NDJSON paths, plus the `--output-format json` invocation contract. 11 tests total now pass.

**MINOR — flagging for your call:** While reviewing `run_claude_review` I noticed the error-message branch at the equivalent of `server.py:284`:

```python
message = str(payload.get("result") or payload.get("error") or result.stderr.strip() or "Claude review failed")
```

doesn't surface `payload.get("errors")` — a list field that recent claude CLI versions emit for budget/rate errors (e.g. `subtype="error_max_budget_usd"`, `is_error=true`, `errors=["Reached maximum budget ..."]`, no `result`/`error` field). Current code degrades to the generic `"Claude review failed"`. This is pre-existing, not introduced by #220, so left out of scope. Happy to either:
- file a follow-up issue and ship #220 as-is, or
- fold a small fix + test into this PR if you'd prefer one round trip.

Whichever you'd rather. Ready for re-review.

treerobin06:
On reflection — folding the MINOR after all in `3514547`. The argument that tipped me: under PR #220's array-parsing patch, what used to surface as a parse error ("did not return JSON output", visible to user) now parses successfully and degrades to the generic "Claude review failed", losing the actionable budget/rate diagnostic. So while the error-message branch itself is pre-existing, this PR makes a previously-noisy failure mode quietly lose information — not strictly out of scope.

5-line change:

```diff
 if result.returncode != 0 or payload.get("is_error"):
+    errors_list = payload.get("errors")
+    errors_text = "; ".join(str(e) for e in errors_list) if isinstance(errors_list, list) and errors_list else ""
     message = str(
         payload.get("result")
         or payload.get("error")
+        or errors_text
         or result.stderr.strip()
         or "Claude review failed"
     )
```

Order preserved so existing `result` / `error` field behavior is unchanged; `errors` list slots in between as a third source before stderr fallback.

Plus 2 mock tests:
- `test_error_result_with_errors_list_surfaces_specific_message`: locks that `"Reached maximum budget"` appears in the returned err and is no longer just `"Claude review failed"`
- `test_error_result_multiple_errors_joined`: locks the `"; "` join semantics

13 tests total now pass:

```
$ python3 -m unittest tests.test_claude_review_parse
Ran 13 tests in 0.103s
OK
```

Branch tip: `3514547`. Ready for re-review.

treerobin06:
One more in `f114726` — internal adversarial pass caught a silent data-loss in the legacy NDJSON fallback path that we missed:

stdout like

```
warning: nvm couldn't find xyz
[{"type":"system",...}, {"type":"result","result":"OK"}]
```

Whole-stdout `json.loads` fails because of the leading non-JSON line. Per-line fallback then scans, parses the array line successfully — but the old branch only accepted `dict` payloads, so the array (with a perfectly good `result` event inside) was dropped, and `parse_claude_json` returned `"Claude CLI did not return JSON output"` despite the result being right there.

Real-world triggers:
- CLI wrappers (nvm, asdf, mise) printing banners to stdout during shell init
- Future `claude --debug` prints
- Any envman that injects "Now using node v20" before exec

Fix: apply the same array-vs-dict policy per-line as the whole-stdout path — when a line parses as `list`, scan reverse for `type=="result"`; if absent, keep scanning earlier lines (don't stop early on a partial array).

Tests:
- `test_noisy_stdout_with_compact_array_line_recovers`: banner + compact array → recovers
- `test_noisy_stdout_with_array_line_no_result_falls_through`: array-line without result → still surfaces the explicit error, no silent grab

15 tests total now pass.

Branch tip: `f114726`. This should round out the parser robustness story; ready for re-review.

wanshuiyin (approved):
Iteration is solid — both blocking points addressed cleanly, plus 290 lines of unit tests covering single-line / multi-line / legacy NDJSON / empty / no-result / noisy banner / errors-list / end-to-end mapping. Bonus fixes for test-ability (stdio extraction) and structured error event surfacing (CLI 2.x errors list) are in-scope and useful. Two minor diagnostic polish items will be tracked in a follow-up issue. Thanks @treerobin06!

PR #224, opened 2026-05-14
fix: tighten parse_claude_json legacy NDJSON diagnostics

### LINKED ISSUE #223: follow-up to #220: tighten parse_claude_json noisy-fallback diagnostics
Follow-up polish from the #220 cross-model review. **Not blocking** — current behavior is safe (no silent success), only diagnostic clarity and one low-probability edge case.

## 1. Noisy-fallback diagnostic regression

`mcp-servers/claude-review/server.py:155-168` — when the legacy per-line NDJSON loop hits a line that parses as a JSON array missing the `result` event, it falls through and keeps scanning earlier lines. If nothing else parses to a usable payload, the final return is the generic `\"Claude CLI did not return JSON output\"` (line 170) rather than the more specific `\"Claude CLI returned a JSON array without a 'result' event\"` that the whole-stdout path emits.

Fix idea: track \`saw_array_without_result = True\` when a list payload misses the result event in the legacy loop, and use that to pick the more specific error message at the end.

## 2. Low-probability silent success in legacy NDJSON path

`mcp-servers/claude-review/server.py:161-162` — within the legacy NDJSON loop, if the input contains BOTH

- a line that parses as a JSON array missing the `result` event, AND
- an earlier line that parses as ANY dict (e.g. a stray `{type: system, ...}` from a CLI 2.x debug print)

then because the loop walks in reverse, it would hit the array first, fall through, then hit the early dict and `return line_payload, None` — silently surfacing a system/init dict as the \"review result\". Same gap that Block #2 closed for the whole-stdout array path, but still open for the per-line fallback.

Real-world likelihood is low (CLI 1.x emits NDJSON only, CLI 2.x emits an array only; mixed shapes are not in either spec). But it is a logic asymmetry with Block #2's intent.

Fix idea: the same \`saw_array_without_result\` flag should also cause the line-level dict path to require `type == \"result\"` (or return the array-without-result error), keeping the two paths symmetric.

## Suggested follow-up test

Add a case `test_legacy_ndjson_with_array_line_without_result_does_not_return_init_dict` covering the scenario in §2.

## Optional (nice-to-have): non-list `errors`

`mcp-servers/claude-review/server.py:296-302` — `errors_list` is surfaced only when it is a non-empty list. If a future CLI emits `errors` as a plain string (e.g. `errors=\"Reached max budget\"`), the diagnostic gets dropped. Trivial to also handle `isinstance(errors_list, str)` and stringify accordingly.

---

Original signal: independent cross-model review on #220 (merged in 745f856). Filing for tracking; happy to take a PR from anyone interested in good-first-issue scope.

algojogacor:
## Summary

Fixes #223

The legacy NDJSON parsing loop in `parse_claude_json` could surface a generic "Claude CLI did not return JSON output" error when a JSON array lacking a `result` event was encountered, instead of the specific "returned a JSON array without a result event" message. Additionally, a low-probability edge case could silently surface a non-result dict (e.g., a system init event) as the review payload.

## Root Cause

In `mcp-servers/claude-review/server.py`, the legacy NDJSON per-line loop (lines 150-167) had no mechanism to track whether it had seen an array-without-result. When scanning lines in reverse:

1. An array line without a `result` event would fall through silently — if no other line yielded a usable payload, the final fallback returned only the generic message (line 167)
2. A dict line appearing earlier in the scan would be returned unconditionally (line 159), even if we had already seen `saw_array_without_result` — potentially surfacing a non-result dict as the review output

## Fix

- Added `saw_array_without_result` flag before the loop, set to `True` when an array line lacks a result event
- In the dict branch: require `type == "result"` when `saw_array_without_result` is set, matching the symmetry the whole-stdout array path already has
- After the loop: return the specific error message when `saw_array_without_result` is set
- Also handle non-list `errors` values (plain strings) in the error extraction path

## Changes

- `mcp-servers/claude-review/server.py`: 12 lines added, 1 line modified (`+12 -1`)

## Testing

- **Legacy NDJSON with array without result, then dict without type=result**: dict is skipped, specific error returned ✅
- **Legacy NDJSON with array without result, no other usable payload**: specific "array without result event" error instead of generic fallback ✅
- **Legacy NDJSON with valid result event in array line (happy path)**: result still returned correctly ✅
- **Whole-stdout path (CLI 2.x)**: unaffected, existing behavior preserved ✅

### DIFF
--- a/mcp-servers/claude-review/server.py
+++ b/mcp-servers/claude-review/server.py
@@ -147,6 +147,7 @@ def parse_claude_json(raw_stdout: str) -> tuple[dict[str, Any] | None, str | Non
     # JSON-array line surrounded by non-JSON noise (wrapper warnings, nvm/asdf
     # banners, future CLI debug prints) still surfaces the result event
     # instead of being silently dropped.
+    saw_array_without_result = False
     for candidate in reversed(stripped.splitlines()):
         candidate = candidate.strip()
         if not candidate:
@@ -156,6 +157,8 @@ def parse_claude_json(raw_stdout: str) -> tuple[dict[str, Any] | None, str | Non
         except json.JSONDecodeError:
             continue
         if isinstance(line_payload, dict):
+            if saw_array_without_result and line_payload.get("type") != "result":
+                continue
             return line_payload, None
         if isinstance(line_payload, list):
             for item in reversed(line_payload):
@@ -163,7 +166,10 @@ def parse_claude_json(raw_stdout: str) -> tuple[dict[str, Any] | None, str | Non
                     return item, None
             # fall through: this line was an array without a result event,
             # but earlier lines might still carry one — keep scanning.
+            saw_array_without_result = True
 
+    if saw_array_without_result:
+        return None, "Claude CLI returned a JSON array without a 'result' event"
     return None, "Claude CLI did not return JSON output"
 
 
@@ -293,7 +299,12 @@ def run_claude_review(
         # Surface it explicitly; otherwise the user sees only the generic
         # "Claude review failed" and loses the actionable message.
         errors_list = payload.get("errors")
-        errors_text = "; ".join(str(e) for e in errors_list) if isinstance(errors_list, list) and errors_list else ""
+        if isinstance(errors_list, list) and errors_list:
+            errors_text = "; ".join(str(e) for e in errors_list)
+        elif isinstance(errors_list, str):
+            errors_text = errors_list
+        else:
+            errors_text = ""
         message = str(
             payload.get("result")
             or payload.get("error")
--- a/tests/test_claude_review_parse.py
+++ b/tests/test_claude_review_parse.py
@@ -140,8 +140,8 @@ class ParseClaudeJsonTests(unittest.TestCase):
         self.assertEqual(payload["result"], "recovered")
         self.assertEqual(payload["session_id"], "sess-noisy")
 
-    def test_noisy_stdout_with_array_line_no_result_falls_through(self) -> None:
-        """Noisy stdout + JSON-array line that has no result event continues scanning earlier lines."""
+    def test_noisy_stdout_with_array_line_no_result_surfaces_specific_diagnostic(self) -> None:
+        """Noisy stdout + JSON-array line with no result event: surface the specific 'array without result event' diagnostic (symmetry with the whole-stdout path)."""
         events_no_result = [_system_init_event(), _assistant_event()]
         stdout = (
             "warning: banner\n"
@@ -149,7 +149,7 @@ class ParseClaudeJsonTests(unittest.TestCase):
         )
         payload, err = MODULE.parse_claude_json(stdout)
         self.assertIsNone(payload)
-        self.assertEqual(err, "Claude CLI did not return JSON output")
+        self.assertEqual(err, "Claude CLI returned a JSON array without a 'result' event")
 
 
 def _completed_process(stdout: str, returncode: int = 0, stderr: str = "") -> subprocess.CompletedProcess:


wanshuiyin (changes requested):
Thanks @algojogacor — clean implementation, all three items from #223 are addressed correctly. One blocking item before merge: the existing test suite from #220 will fail under the new error message.

## Blocking: existing test expects the old generic error

`tests/test_claude_review_parse.py:143-152` covers exactly the scenario this PR changes:

```python
def test_noisy_stdout_with_array_line_no_result_falls_through(self) -> None:
    \"\"\"Noisy stdout + JSON-array line that has no result event continues scanning earlier lines.\"\"\"
    events_no_result = [_system_init_event(), _assistant_event()]
    stdout = (
        \"warning: banner\\n\"
        + json.dumps(events_no_result) + \"\\n\"
    )
    payload, err = MODULE.parse_claude_json(stdout)
    self.assertIsNone(payload)
    self.assertEqual(err, \"Claude CLI did not return JSON output\")  # ← will now be the array-without-result error
```

After your patch, the same input now sets `saw_array_without_result=True` and returns `\"Claude CLI returned a JSON array without a 'result' event\"`. Please update this test's expected error string to match.

## Nice-to-have new tests (not blocking but recommended)

The three behaviors your patch introduces deserve their own test cases. Adding these to `tests/test_claude_review_parse.py` would make the regression coverage robust:

1. **Legacy NDJSON: array-without-result line followed by non-result dict line is skipped + specific error returned.**
   ```python
   def test_legacy_ndjson_array_then_init_dict_returns_specific_error(self) -> None:
       \"\"\"Source order: [array-no-result, then init dict line]. Reverse scan should set the flag on the array and skip the init dict.\"\"\"
       stdout = (
           json.dumps([_system_init_event(), _assistant_event()]) + \"\\n\"
           + json.dumps(_system_init_event()) + \"\\n\"
       )
       payload, err = MODULE.parse_claude_json(stdout)
       self.assertIsNone(payload)
       self.assertEqual(err, \"Claude CLI returned a JSON array without a 'result' event\")
   ```

2. **Legacy NDJSON: array-with-result still wins even after seeing a non-result dict.**
   Verifies your existing happy-path regression — the symmetry should NOT eat a valid result event when one is present.

3. **`run_claude_review`: string-form `errors` field is surfaced.**
   Covers the new \`elif isinstance(errors_list, str)\` branch you added in `run_claude_review`.

## Implementation review (passes)

- Item 1 (`saw_array_without_result` flag + specific error at end): `server.py:150, 169-172`. ✓
- Item 2 (symmetry: skip non-result dicts after flag set): `server.py:159-162`. ✓ Matches the whole-stdout array path's `\"type\"==\"result\"` requirement.
- Item 3 (non-list `errors` handling): `server.py:301-307`. ✓ Also degrades gracefully on `None` / non-string non-list (falls through to stderr/generic).

Reverse-scan edge case I traced (`[array-with-result, dict-without-result]`): the dict-without-result is processed first in the reversed scan, no flag set yet, so it returns the dict. This is a theoretical case for CLI 1.x mixed output that the spec doesn't actually produce — not a real blocker. If you want to harden it further (preferring `type==\"result\"` over typed non-result dicts globally), that's a Round 2 issue, not this PR.

Will merge once the existing test is updated. Thanks for the careful work on this.

wanshuiyin:
Thanks for the clean fix and the excellent root-cause writeup — this is exactly the kind of PR I love to merge.

Applied locally: 14/15 existing tests pass. The single failure is `tests/test_claude_review_parse.py:143` (`test_noisy_stdout_with_array_line_no_result_falls_through`) — that test was still asserting the OLD generic fallback for the noisy-stdout + array-without-result case. Your fix actually does the more correct thing here: the legacy NDJSON path now surfaces the same specific diagnostic that the whole-stdout path already exposes via `test_array_without_result_event_returns_error`.

Since the test update is a single string change and maintainer edits are enabled, I'll push that one-line update to this PR branch as a separate maintainer commit, so we don't bounce a tiny follow-up back and forth. Merging right after CI is green.

Thanks again — closes #223 cleanly.

PR #231, opened 2026-05-18
Add Codex OAuth backend to llm-chat

angziii:


wanshuiyin:
Thanks for this — really clean implementation. The design is sound: read-only sandbox + ephemeral + disabled plugins + list-argv subprocess gives a solid security floor, and the `auto` fallback to codex-cli when no API key is set is genuinely useful for users on ChatGPT-only plans.

I reviewed the diff carefully + ran the 26 tests locally (all pass). I think we should cover the items below before merging — happy to clarify any of them.

## Blockers

**1. Test coverage on the security-critical subprocess command**

The current 3 tests cover happy path / `auto` fallback / non-zero exit. For a backend that may run with user auth/config, please also add assertions that the command actually includes the safety flags. Suggested additions in `tests/test_llm_chat_server.py`:

- Assert `--sandbox`, `read-only`, `--ephemeral`, `--disable`, `plugins`, `-C`, and `-` (stdin marker) are all in the cmd list. Right now only `--ephemeral` and `--output-last-message` are asserted; the rest could regress in a future refactor.
- New test: `subprocess.TimeoutExpired` branch — raise `TimeoutExpired` from mock, assert error contains `"timed out"`.
- New test: `FileNotFoundError` branch — covers misconfigured `CODEX_BIN`. Should surface `"Codex CLI not found: <bin>"`.

**2. README threat-model note**

Two specific lines worth adding to the Codex OAuth section:

- "`codex-cli` is an **agent backend**, not a pure chat API. `codex exec` loads local Codex config, rules, and `AGENTS.md` files — only session state is killed by `--ephemeral`. For reviewer-independence guarantees in cross-model review, point `CODEX_WORKDIR` at a narrow, dedicated directory rather than the project root, and consider absolute paths for `CODEX_BIN`."
- "Even with `--sandbox read-only`, Codex has filesystem read access to `CODEX_WORKDIR` — set it deliberately."

This protects users from accidentally exposing more context than they intended, especially since `CODEX_WORKDIR` defaults to `os.getcwd()`.

## Merge hygiene

- This is an auth/backend PR — a short body covering "what this adds / what could break / how I tested it" really helps future readers, especially when we surface this in news/changelog. 3-5 lines is enough.

## Nice to have (not blocking)

- `CODEX_TIMEOUT_SECS = int(os.environ.get(...))` will crash the server on a non-numeric value. A `try/except ValueError` with fallback to 600s would be friendlier.
- The stdin format `ROLE:\ncontent` joined by blank lines is OK for a single reviewer turn, but a user message containing literal `SYSTEM:\n...` text could spoof a role boundary. Worth a one-line caveat in the README.
- Small test asserting the tmp output path doesn't exist after the call (cleanup regression guard).

## What I'll handle on the merge commit

ARIS news entry on the merge commit + categorization in the LLM_API_MIX_MATCH_GUIDE — no action needed from you there.

Happy to take the `try/except ValueError` and the README threat-model note off your plate if that's easier — just say the word. The test additions I'd really like to come from you since you wrote the surrounding pattern, but happy to pair if helpful.

angziii:
Thanks for the detailed review. I pushed a follow-up commit to this PR addressing the blockers and the small robustness items:

- Expanded Codex CLI backend tests to assert `--disable plugins`, `--sandbox read-only`, `--ephemeral`, `-C`, stdin `-`, timeout propagation, and output-file cleanup.
- Added tests for `FileNotFoundError` and `subprocess.TimeoutExpired` branches.
- Added README threat-model notes for `codex-cli` as an agent backend, `CODEX_WORKDIR` read exposure, and prompt role-header caveat.
- Made `CODEX_TIMEOUT_SECS` parsing tolerant of invalid/non-positive values.

Validation: `py_compile`, `tests.test_llm_chat_server` now 29 tests, and `git diff --check` all pass.

wanshuiyin (changes requested):
Thanks for this — a no-API-key reviewer fallback through the local Codex login is genuinely useful, and defaulting `LLM_BACKEND=api` (opt-in, non-breaking) is the right call. A cross-model review (Claude + GPT-5.5 xhigh) surfaced a few things to harden before merge, plus a rebase. Direction is good; these are about the exec surface.

## 🔴 P1 — scrub the child environment (secret-exfil surface)
`mcp-servers/llm-chat/server.py` runs `codex exec` via `subprocess.run` without a sanitized `env=`, so the child inherits the full MCP-server environment + the user's Codex config. The reviewer prompt is attacker-influenceable content (you're reviewing arbitrary research text / PRs), so a crafted prompt can induce the Codex agent to read and print env secrets (`LLM_API_KEY`, `OPENAI_API_KEY`, `*_TOKEN`) — which then flow back out through the echoed output. `--sandbox read-only` blocks writes but not reading/printing env.
→ Pass a minimal whitelisted `env=`, and use a dedicated `CODEX_HOME` so local rules/config aren't pulled into a reviewer call.

## 🟡 P2 — don't echo raw CLI output as MCP content/errors
The path returns raw `stderr`/`stdout` (up to 1000 chars) and `str(e)` on failure — that can leak transcripts, file paths, and config details. Return generic auth/timeout/exec errors and redact known secret patterns.

## 🟡 P2 — `LLM_BACKEND=auto` silently switches to `codex-cli` when `LLM_API_KEY` is absent
Not unauthenticated, but after a missing-key misconfig it silently routes reviewer prompts through the local ChatGPT/OAuth Codex account. For a security-sensitive review path, require an explicit `LLM_BACKEND=codex-cli`, or make `auto` warn loudly before falling through.

## 🟡 P2 — tests exercise a copied helper, not the server
The added tests patch `tests/_llm_chat_helpers`, while the production logic lives in `mcp-servers/llm-chat/server.py`; they also don't assert env scrubbing, output redaction, or real auth-failure behavior. Please exercise the actual code path.

## Rebase
The diff no longer applies cleanly to current `main` (`mcp-servers/llm-chat/server.py` diverged — `import datetime` + a datetime debug_log). A 3-way merge resolves with no semantic conflict, so a rebase is all that's needed.

**Reviewer-independence note (non-blocking):** the codex-cli backend is the *same GPT family* as the existing Codex MCP reviewer. Fine as a no-key fallback, but it must not be used as the cross-model jury when the executor is also GPT — worth a sentence in the guide you added.

Happy to re-review as soon as the env scrubbing + output redaction land. Thanks again for the contribution!

PR #242, opened 2026-05-22
Fix Windows PowerShell installer parity

VeraPyuyi:
﻿## Summary
- replace the Windows PowerShell nested junction installer with flat per-skill junction installs for Claude and Codex
- add manifest-backed apply/reconcile/uninstall, dry-run, safe conflict reporting, per-skill `-ReplaceLink`, and stale-lock cleanup
- update managed `CLAUDE.md` / `AGENTS.md` blocks to reference flat skill paths and refresh README Windows examples

## Safety
- real files and user-owned skill directories are never overwritten
- uninstall only removes junctions/symlinks recorded in the relevant manifest
- dry-run paths avoid creating `.aris`, `.claude`, `.agents`, or docs

## Tests
- `python -m pytest tests/test_install_aris_ps1.py -q`
- `powershell -NoProfile -ExecutionPolicy Bypass -File tools/install_aris.ps1 . -Platform codex -ArisRepo . -DryRun`
- `git diff --check`


wanshuiyin:
Thanks @VeraPyuyi for tackling Windows parity — the nested-junction layout has been a known sharp edge. Fresh-thread cross-model review (codex gpt-5.5 xhigh, no shared context) found **two blockers** + parity gaps with the Unix installer. The PowerShell logic itself is mostly sound, but a few things need fixing before merge:

---

### 🔴 1. README.md / README_CN.md changes are pure CRLF conversion noise

The huge 2105+/2104- per file is **not real content changes** — it's your editor (or `git config core.autocrlf true` on Windows) converting LF to CRLF. Verified with `od -c`: original lines end `\n`, your `+` lines end `\r\n`. The actual repo convention is LF (current README.md is LF, no `text=auto` in `.gitattributes`).

Merging this would dirty the entire repo history with line-ending noise.

**Fix**:
```bash
# In your branch
git config core.autocrlf input
git checkout main -- README.md README_CN.md   # restore originals
git commit --amend
git push --force-with-lease
```

Or add `* text=auto eol=lf` to `.gitattributes` first to lock down for the future.

### 🔴 2. `UPDATE_TARGET` silently overwrites user-customized junctions

This contradicts the PR's own safety claim ("user-owned skill directories are never overwritten"). Concrete scenario:

1. User installs ARIS once → junction `~/.claude/skills/research-lit` → `<ARIS>/skills/research-lit`
2. User wants their own variant → manually re-points the junction to `~/my-fork/skills/research-lit`
3. User re-runs the installer to pick up new skills
4. `Compute-Plan` (install_aris.ps1:296-298) sees in-manifest entry with "wrong" target → marks as `UPDATE_TARGET`
5. `Apply-Plan` (install_aris.ps1:458-463) `Remove-LinkPath` + `New-Junction` → user's customization is silently destroyed

The Unix installer guards against this by checking whether the current target is inside the ARIS repo before deciding to overwrite. The PS rewrite needs equivalent logic — e.g., only `UPDATE_TARGET` if `currentTarget` resolves under `$ArisRepo`; otherwise warn + skip.

### 🟡 3. Missing legacy nested-junction migration

`tools/install_aris.sh:309-353` detects the old nested layout (`.claude/skills/aris/<skill>`) and requires explicit `--from-old` / `--migrate-copy`. The PS rewrite has no equivalent — users with the old layout will end up with both nested junctions AND new flat junctions, both pointing into ARIS, with the doc managed-block listing only the new ones. Confusing state.

### 🟡 4. Missing `.aris/tools` symlink

`tools/install_aris.sh:533-568` (`ensure_tools_symlink`) creates `.aris/tools` → `<ARIS>/tools` so skills can find helpers via the canonical 3-layer resolver (`.aris/tools/<helper>` first). Without this, every cross-skill helper invocation on Windows falls through to layer-2 (`tools/<helper>` relative to cwd) or layer-3 ($ARIS_REPO env var), neither of which is guaranteed in the target project.

### 🟡 5. Tests effectively never run

`tests/test_install_aris_ps1.py:14-17` skips entirely if `powershell` is missing. Two issues:
- No Windows CI job in the repo → these tests never run on PR
- Skip predicate only checks `powershell`, not `pwsh` → PowerShell 7+ on macOS/Linux is excluded even though the script claims `#Requires -Version 5.1` would work there

Suggest: (a) add a GitHub Actions Windows job that runs this test suite, AND (b) try `pwsh` first, fall back to `powershell.exe` — so cross-platform contributors can run the tests locally.

---

✅ **Things that look good**:
- `-DryRun` properly returns before mkdir/lock/doc/manifest writes (`install_aris.ps1:682-685`)
- Real-directory conflict aborts instead of overwriting (`install_aris.ps1:303-306`)
- Tests do invoke real PowerShell + verify junctions (`tests/test_install_aris_ps1.py:20-28, 98-102`)
- Lock cleanup logic is sound for its scope (just `.aris/.install*.lock.d`)

Once #1 (CRLF) and #2 (user-junction guard) are fixed, this is mergeable. #3-5 can be follow-ups (the PR description already notes parity is a work-in-progress).

Codex thread: `019e5ab5-e505-7103-a153-dd414957e674`.

VeraPyuyi:
Addressed the review items in `a473994`:

- rebased onto current `origin/main` and force-updated the branch through normal git, replacing the earlier API-published head; GitHub now reports the PR as mergeable
- added `.gitattributes` with LF normalization and confirmed `README.md` / `README_CN.md` are small LF-only diffs
- changed `UPDATE_TARGET` so managed junctions are only relinked when the current target is still inside the configured ARIS repo; user-retargeted external junctions now stop as conflicts and remain untouched, even with `-ReplaceLink`
- added legacy nested install handling with `-FromOld` and `-MigrateCopy keep-user|prefer-upstream`
- added managed `.aris\tools` junction creation and exact-target uninstall cleanup
- added Windows GitHub Actions coverage for `tests/test_install_aris_ps1.py`
- updated tests to prefer `pwsh` and fall back to Windows PowerShell; tests remain Windows-only because they verify Windows junction behavior

Validation run locally:

```text
python -m pytest tests/test_install_aris_ps1.py -q  # 7 passed
powershell -NoProfile -ExecutionPolicy Bypass -File tools\install_aris.ps1 . -Platform codex -ArisRepo . -DryRun
git diff --check
line-ending scan for .gitattributes / README.md / README_CN.md / tools/install_aris.ps1 / tests/test_install_aris_ps1.py / workflow  # CRLF=0 for all
```

VeraPyuyi:
Follow-up review pass addressed the Windows installer parity findings:

- Added S10-style source safety: upstream skill/source junctions resolving outside the configured ARIS repo are skipped and never installed into project skills.
- CI now runs the installer tests in a matrix for \\powershell.exe\\ and \\pwsh\\; tests also accept \\ARIS_TEST_POWERSHELL\\ for explicit runner coverage.
- Strengthened the stale junction regression so it uses PowerShell \\Get-Item -LiteralPath -Force\\, not \\Path.exists()\\, to prove a broken junction reparse point was removed.
- Fixed shared \\.aris/tools\\ uninstall behavior: uninstalling one platform keeps the tools junction when the other platform manifest still points at the same ARIS repo.

Validation:
- \\python -m pytest tests/test_install_aris_ps1.py -q\\ -> 9 passed
- \\ARIS_TEST_POWERSHELL=powershell.exe python -m pytest tests/test_install_aris_ps1.py -q\\ -> 9 passed
- Codex dry-run against this repo -> 69 CREATE / 0 CONFLICT, no writes
- \\git diff --check\\ -> clean

wanshuiyin:
Thanks @VeraPyuyi — exceptional follow-up. All 5 items from round 1 cleanly fixed, plus you proactively added `.gitattributes` LF normalization and a Windows CI matrix for both `pwsh` and `powershell.exe`. The `Test-PathInside` guard correctly handles Windows case-insensitivity, and `-MigrateCopy keep-user|prefer-upstream` is a nice touch.

Ran a second-pass codex xhigh on the latest tree — **all core findings PASS**. One corner-case observation, not a merge blocker:

---

### ⚪ Corner case: `Test-PathInside` only resolves one hop

`Test-PathInside $currentTarget $Config.RepoRoot` checks whether the junction's first-hop target is inside the ARIS repo, but doesn't follow further indirection. Scenario:

1. User creates `<ARIS>/skills/research-lit` as a symlink to `~/my-fork/skills/research-lit` (advanced customization — overlay their fork onto the canonical ARIS tree)
2. `.claude/skills/research-lit` junction → `<ARIS>/skills/research-lit` (inside-repo by first-hop test)
3. Guard reports `UPDATE_TARGET` as safe, relinks → user's customization layer is silently bypassed on next install

This is a corner-of-corner case (most users who customize point their junctions directly outside ARIS, which IS correctly guarded), and the existing fix handles 99% of real cases. But the indirect-symlink chain bypasses it.

**Options**:
- **(a)** Use `Resolve-Path -LiteralPath` in a loop to follow the full reparse chain before deciding inside/outside
- **(b)** Document the limitation in the PR description and track as known follow-up
- **(c)** Ship as-is — the guard catches the realistic attack surface

Personally I'd lean (c) + (b): the realistic threat is "user accidentally re-points their own junction outside ARIS, installer overwrites it" — that's solidly fixed. The symlink-chain case requires intentional advanced setup that already has 3-layer indirection; users doing that should know the model.

Your call. If you go (c), we can merge this now and pick up (a) in a follow-up PR if it becomes a real user issue. If you go (a), it's a small loop change.

Codex thread: `019e5ddb-2ee0-7f02-a642-83ecafdd527e`.

VeraPyuyi:
Addressed the second-pass corner case by taking option (a), then rebased the PR onto current `origin/main` after `/wiki-enrich` landed upstream.

Changes now pushed at `154d7f2`:
- Added `Resolve-ReparseChain` and `Test-ResolvedPathInside` so installer safety checks follow the full junction/symlink chain before deciding whether a target is inside the configured ARIS repo.
- Updated upstream source-link filtering plus `UPDATE_TARGET` planning/apply guards to use the resolved-chain check.
- Added a regression where `.agents/skills/alpha` first points inside ARIS, but that inside target is itself a junction to an external source; even with `-ReplaceLink alpha`, the installer now reports `CONFLICT` and leaves the user target untouched.
- Rebasing onto latest `main` pulled in `/wiki-enrich`, so this branch also syncs its Codex mirror/catalog/counts and installs `pytest` in the inventory workflow before running mirror semantics tests.

Validation:
- `python tools/check_skills_inventory.py` -> passed
- `python -m pytest tests/test_codex_skill_mirror.py -q` -> 16 passed
- `python -m pytest tests/test_install_aris_ps1.py -q` -> 10 passed
- `ARIS_TEST_POWERSHELL=powershell.exe python -m pytest tests/test_install_aris_ps1.py -q` -> 10 passed
- Codex dry-run against this repo -> 78 CREATE / 0 CONFLICT, no writes
- `git diff --check` -> clean

GitHub checks on PR #242 are now all green: inventory check, advisory lint, `test (pwsh)`, and `test (powershell.exe)`.

VeraPyuyi:
Thanks again for the careful review. I did one more pass specifically on the indirect junction case and found a narrower variant where a parent path component, not the leaf, was the reparse point.

Follow-up fix pushed in `3b591bd`:
- `Resolve-ReparseChain` now walks path components from root to leaf and rewrites at the first junction/symlink component, so intermediate reparse points are resolved before inside-repo checks.
- `Build-Inventory` now validates every candidate skill's resolved path against the resolved ARIS repo root, not only leaf entries that are themselves links.
- Added regressions for an external source root junction and a manifest-retargeted skill pointing through an in-repo parent junction.

Validation:
- `python -m pytest tests/test_install_aris_ps1.py -q` -> 12 passed
- `$env:ARIS_TEST_POWERSHELL='powershell.exe'; python -m pytest tests/test_install_aris_ps1.py -q; Remove-Item Env:\ARIS_TEST_POWERSHELL` -> 12 passed
- `python tools/check_skills_inventory.py` -> passed
- `python -m pytest tests/test_codex_skill_mirror.py -q` -> 16 passed
- `git diff --check` -> clean

GitHub checks are green on the new head as well.

PR #245, opened 2026-05-23
fix(claude-review): Windows compatibility for subprocess spawning

### LINKED ISSUE #244: fix(claude-review): server.py crashes on Windows due to Unix-only subprocess patterns
## Problem

`mcp-servers/claude-review/server.py` uses several Unix-only patterns that cause crashes or incorrect behavior on Windows:

### 1. `start_new_session=True` (line 368)
Windows does not support `start_new_session`. This causes `OSError` or unexpected behavior when spawning the Claude CLI worker process.

### 2. `os.kill(pid, 0)` for liveness check (line 199)
On Windows, `os.kill(pid, 0)` does NOT check liveness — it **terminates** the process (sends `CTRL_BREAK_EVENT` or calls `TerminateProcess`). This kills the worker immediately after spawning it.

### 3. Hardcoded `/tmp/` path (line 41)
Windows has no `/tmp/` directory. The debug log path fails silently or errors.

## Impact

Codex → Claude review workflow (Topology B) is completely broken on Windows. The worker process is either killed immediately after spawn or fails to start.

## Fix

Platform-conditional logic:

```python
# 1. Process creation
if sys.platform == "win32":
    popen_kwargs["creationflags"] = subprocess.CREATE_NEW_PROCESS_GROUP
else:
    popen_kwargs["start_new_session"] = True

# 2. Liveness check
if sys.platform == "win32":
    import ctypes
    kernel32 = ctypes.windll.kernel32
    SYNCHRONIZE = 0x00100000
    handle = kernel32.OpenProcess(SYNCHRONIZE, False, pid)
    if handle:
        kernel32.CloseHandle(handle)
        return True
    return False
else:
    os.kill(pid, 0)

# 3. Temp path
_default_debug_dir = Path(os.environ.get("TEMP", "/tmp"))
```

Additional improvements in the fix:
- Worker stdout/stderr redirected to log files (not DEVNULL) for debuggability
- File handles passed to Popen closed in `finally` block (Windows file locking is strict)
- `cwd=os.getcwd()` passed explicitly to ensure worker inherits correct directory

## Environment

- Windows 11 Pro
- Python 3.12
- Codex CLI (latest)
- Claude Code CLI 2.x

Will submit a PR with the complete fix.

King-ty:
## Summary

- Fix `start_new_session=True` crash on Windows → use `CREATE_NEW_PROCESS_GROUP`
- Fix `os.kill(pid, 0)` killing the worker on Windows → use `OpenProcess(SYNCHRONIZE)`
- Fix hardcoded `/tmp/` path → use `TEMP` env var on Windows
- Redirect worker stdout/stderr to log files for debuggability
- Close file handles in `finally` block (Windows file locking)
- Pass explicit `cwd` to Popen

## Context

The `claude-review` MCP server (Topology B: Codex executor → Claude reviewer) is completely broken on Windows. The worker process either fails to start or gets immediately killed by the liveness check.

Fixes #244

## Test plan

- [x] Existing 15 unit tests pass (`tests/test_claude_review_parse.py`)
- [x] Manual smoke test: Codex calls `review_start` → polls `review_status` → gets `CLAUDE_REVIEWER_READY` response on Windows 11
- [ ] Verify no regression on Linux/macOS (all changes are platform-conditional)

### DIFF
--- a/mcp-servers/claude-review/server.py
+++ b/mcp-servers/claude-review/server.py
@@ -38,7 +38,8 @@ DEFAULT_MODEL = os.environ.get("CLAUDE_REVIEW_MODEL", "")
 DEFAULT_SYSTEM = os.environ.get("CLAUDE_REVIEW_SYSTEM", "")
 DEFAULT_TOOLS = os.environ.get("CLAUDE_REVIEW_TOOLS", "")
 DEFAULT_TIMEOUT_SEC = int(os.environ.get("CLAUDE_REVIEW_TIMEOUT_SEC", "600"))
-DEBUG_LOG = Path(os.environ.get("CLAUDE_REVIEW_DEBUG_LOG", f"/tmp/{SERVER_NAME}-mcp-debug.log"))
+_default_debug_dir = os.environ.get("TEMP", "/tmp") if sys.platform == "win32" else "/tmp"
+DEBUG_LOG = Path(os.environ.get("CLAUDE_REVIEW_DEBUG_LOG", f"{_default_debug_dir}/{SERVER_NAME}-mcp-debug.log"))
 STATE_DIR = Path(
     os.environ.get(
         "CLAUDE_REVIEW_STATE_DIR",
@@ -194,6 +195,15 @@ def job_state_path(job_id: str) -> Path:
 
 def is_pid_alive(pid: int | None) -> bool:
     if not pid or pid <= 0:
+        return False
+    if sys.platform == "win32":
+        import ctypes
+        kernel32 = ctypes.windll.kernel32
+        SYNCHRONIZE = 0x00100000
+        handle = kernel32.OpenProcess(SYNCHRONIZE, False, pid)
+        if handle:
+            kernel32.CloseHandle(handle)
+            return True
         return False
     try:
         os.kill(pid, 0)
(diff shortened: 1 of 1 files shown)


wanshuiyin:
Thanks for the careful Windows fix — the 3 distinct bugs (`start_new_session=True` crash, `os.kill(pid, 0)` terminating the worker, hardcoded `/tmp/`) are real, and the test plan with the Windows 11 smoke test is exactly the right way to validate.

Read through the diff:

- ✅ Changes 1 (DEBUG_LOG default) and 2 (is_pid_alive Windows branch) are pure additions behind `sys.platform == "win32"` guards — zero impact on macOS/Linux.
- ✅ Change 3 (Popen kwargs) is the only cross-platform shift: worker stdout/stderr now go to `<JOBS_DIR>/<job_id>.worker.{out,err}.log` files instead of DEVNULL. Net upside is debuggability on every platform; minor downside is log files accumulate in JOBS_DIR (no cleanup), but that mirrors the existing pattern for job state JSONs so it's consistent.
- ✅ Finally-block file close is correct (child process keeps its own fd copies after parent closes the parent's handles).

Fixes #244, no other touchpoints. Merging.

PR #247, opened 2026-05-24
feat(wiki-enrich): add skill to fill paper TODO sections left by ingest_paper

hungchun0201:
## Summary

Add `/wiki-enrich`, a new community skill that fills the per-paper TODO sections that `ingest_paper` leaves behind.

## Why this skill

Every paper-reading skill (`/research-lit`, `/arxiv`, `/alphaxiv`, `/deepxiv`, `/semantic-scholar`, `/exa-search`) ends by calling `ingest_paper`, which renders a scaffold with frontmatter, abstract, and 12 `_TODO._` placeholder sections. No downstream skill ever fills those sections — `tools/research_wiki.py:694` even acknowledges this in the `--thesis` help text: *"otherwise left as TODO for later enrichment"*, but the enrichment mechanism was never implemented.

This contradicts the Karpathy LLM-wiki design that ARIS's `/research-wiki` cites in its acknowledgements:

> *"You never (or rarely) write the wiki yourself — the LLM writes and maintains all of it. … LLMs don't get bored, don't forget to update a cross-reference, and can touch 15 files in one pass."*
> — [karpathy/442a6bf555914893e9891c11519de94f](https://gist.github.com/karpathy/442a6bf555914893e9891c11519de94f)

In practice, after a normal `/research-lit` run the wiki sits as TODO until someone reads each paper by hand — exactly the maintenance burden Karpathy warns kills wikis.

## What `/wiki-enrich` does

For each paper page with TODO sections:

1. Reads the YAML frontmatter to extract `arxiv` id and `title`.
2. Fetches paper content via a graceful fallback chain: **alphaxiv overview → alphaxiv abs → deepxiv brief → arXiv API abstract → existing page abstract**.
3. Generates 1–3 sentence prose for each TODO section (One-line thesis, Problem / Gap, Method, Key Results, Assumptions, Limitations, Reusable Ingredients, Open Questions, Claims, Relevance to This Project).
4. Edits the page in place via the `Edit` tool, replacing each `_TODO._` block.
5. Appends an audit-trail entry to `log.md` recording which source provided the content.

`## Connections` and `## Abstract (original)` are never touched.

## Targets

- `/wiki-enrich` — default: every paper with ≥1 TODO (recommended after batch ingest)
- `/wiki-enrich <slug>` — single paper
- `/wiki-enrich all --force` — rewrite every paper from scratch
- `--source alphaxiv|deepxiv|arxiv|auto`, `--max N` — knobs for token budget / source pinning

## Additive only

- No changes to existing skills.
- No changes to `tools/research_wiki.py` (uses the existing `log` subcommand for the audit trail).
- Any ingest pipeline can optionally append `/wiki-enrich` afterward.

## Files

- `skills/wiki-enrich/SKILL.md` (~245 lines)

### DIFF
--- a/skills/wiki-enrich/SKILL.md
+++ b/skills/wiki-enrich/SKILL.md
@@ -1,0 +1,245 @@
+---
+name: wiki-enrich
+description: "Fill in the per-paper TODO sections (Problem/Method/Key Results/Limitations/Reusable Ingredients/...) of research-wiki/papers/<slug>.md pages that /research-lit, /arxiv, /alphaxiv, /deepxiv, /semantic-scholar, /exa-search ingest as bare scaffolds. Implements the Karpathy LLM-wiki principle that the LLM (not the human) writes and maintains wiki bodies. Use when user says 'enrich wiki', 'fill paper TODOs', 'wiki body 補完', '把 paper 摘要寫進 wiki', 'research-wiki 自動填', or after a batch ingest that left papers/ as TODO scaffolds."
+argument-hint: [target: slug|missing|all] [--source alphaxiv|deepxiv|arxiv|auto] [--force] [--max N]
+allowed-tools: Bash(*), Read, Write, Edit, Glob, Grep, WebFetch
+---
+
+# Wiki Enrich: Fill Paper TODO Sections (Karpathy LLM-Wiki)
+
+Target: **$ARGUMENTS**
+
+## Why this skill exists
+
+`ingest_paper` (called by `/research-lit`, `/arxiv`, `/alphaxiv`, `/deepxiv`, `/semantic-scholar`, `/exa-search`) only renders the per-paper scaffold — frontmatter + abstract + 12 `_TODO._` placeholder sections. No downstream skill in ARIS fills those sections; the wiki sits as TODO until someone reads each paper.
+
+This contradicts the Karpathy LLM-wiki design (https://gist.github.com/karpathy/442a6bf555914893e9891c11519de94f):
+
+> "You never (or rarely) write the wiki yourself — the LLM writes and maintains all of it. … The tedious part of maintaining a knowledge base is not the reading or the thinking — it's the bookkeeping. … LLMs don't get bored, don't forget to update a cross-reference, and can touch 15 files in one pass."
+
+`/wiki-enrich` is the missing back half of `ingest_paper`: it reads each scaffolded paper page, fetches paper content from external sources (alphaxiv overview → deepxiv brief → arXiv abstract), and rewrites the 12 TODO sections into 1-3 sentence prose summaries.
+
+## Constants
+
+- **WIKI_ROOT = `research-wiki/`** — Resolved relative to git root. Skill hard-fails if not a directory.
+- **TARGET_DEFAULT = `missing`** — When no target is given, enrich only papers with ≥1 TODO section. Other targets: `<slug>` (one paper) or `all` (every paper, even ones already enriched — usually combined with `--force` to overwrite).
+- **SOURCE_DEFAULT = `auto`** — Fetch order: alphaxiv overview → deepxiv brief → arXiv API abstract. First non-empty wins. Override with `--source` to pin one source.
+- **MAX_PAPERS = 20** — Hard cap per invocation; LLMs touch many files but token budgets are real. Override with `--max N`.
+- **FORCE = false** — When `false` (default), skip sections that already have non-TODO content. When `true`, overwrite every section except `## Connections` (always auto-generated from edges.jsonl, never written by hand).
+- **SECTIONS_TO_FILL** — The 12 sections `ingest_paper` scaffolds:
+  1. `One-line thesis` (marker: `_TODO: fill in after reading._`)
+  2. `Problem / Gap` (marker: `_TODO._`)
+  3. `Method` (marker: `_TODO._`)
+  4. `Key Results` (marker: `_TODO._`)
+  5. `Assumptions` (marker: `_TODO._`)
+  6. `Limitations / Failure Modes` (marker: `_TODO._`)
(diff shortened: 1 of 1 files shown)


wanshuiyin:
Thanks @hungchun0201 — this is a real gap. `ingest_paper` has been leaving TODO scaffolds across the codebase and nothing was picking them up, exactly the Karpathy maintenance-debt failure mode. Direction 👍.

Did a fresh-thread cross-model review (codex gpt-5.5 xhigh, no shared context). One real bug + a few documentation/precision items:

---

### 🔴 1. Phase 1 regex silently misses the One-line thesis marker

`research_wiki.py:437` emits literally:

```
_TODO: fill in after reading._
```

(trailing **`._`**, dot + underscore). But Phase 1's grep is:

```bash
grep -lE "^_TODO(\._?|: fill in after reading\.)$" research-wiki/papers/*.md
```

The second alternative ends with `\.` not `\._`, so papers where the One-line thesis is the **only** remaining TODO get silently dropped from the `missing` target. Easy fix:

```bash
grep -lE "^_TODO(\._?|: fill in after reading\._?)$" research-wiki/papers/*.md
```

(or just `^_TODO[._: a-z]+$` — looser but simpler).

### 🟡 2. "12 TODO sections" is overstated — actually 10 fillable

PR description, SKILL line 14, line 20, and the Constants enumeration all say *"12 `_TODO._` placeholder sections"*. But `ingest_paper` only writes `_TODO._` on **10** sections (`research_wiki.py:436-467`):
- `## Connections` (#10) is a graph-summary placeholder, not a TODO
- `## Abstract (original)` (#12) is either populated by `ingest_paper` from arXiv API or omitted entirely

SKILL itself already says both must NEVER be edited — so the "12" count contradicts its own table. Suggest changing the headline number to **10** everywhere.

### 🟡 3. Audit-trail label drift

Step 2.6 emits source as `page-abstract-fallback`; final report (Phase 3) lists it as `page-fallback`. Same source, two names — pick one so log grepping works.

### 🟡 4. arXiv row conflates two different sources

Phase 2.3 table row 4 covers both "abstract already in the page body" AND "fresh curl from arXiv API". These are different provenance (one is stale-from-ingest, one is fresh-from-network), and the audit log can't distinguish them downstream. Suggest splitting into two rows + two log labels (`arxiv-api-abstract` vs `page-abstract-fallback`).

### ⚪ 5. Resolver chain uses pre-Arch-C form (not blocking)

Pre-flight uses the older bare-form `ARIS_REPO="${ARIS_REPO:-$(awk ...)}"` not the strict-safe `if [ -z ... ] && [ -f ... ]` variant from `shared-references/integration-contract.md` §2. This matches what `/idea-creator` SKILL still does today, so the PR is consistent with the existing precedent — not a blocker. But if you want to be ahead of the curve, copy the §2 strict-safe block (handles `set -e` callers cleanly).

---

🔴 #1 is the only must-fix. The rest are quick documentation tightenings. Once #1 is in I think this can merge — really nice addition, the Karpathy framing in the PR description is also a good piece of design rationale we should preserve in the SKILL itself.

Codex thread for transparency: `019e5a83-9c6f-7291-a501-7b40c1a3f324`.

hungchun0201:
Thanks for the careful review @wanshuiyin — really appreciate the codex cross-thread audit, the regex bug was a real one. Round-2 commit `fb76f6d` pushed; all four items addressed:

| # | Action |
|---|---|
| 🔴 1. regex | `\.)$` → `\._?)$`. Verified against synthetic page that had **only** the One-line thesis TODO — old regex missed it, new one matches. |
| 🟡 2. "12 → 10" | Reworded everywhere ("Why this skill", `SECTIONS_TO_FILL`, intent line). Enumeration kept as-is (still 12 entries since the section headers exist); description now correctly says 10 fillable + 2 protected (positions 10 and 12). |
| 🟡 3. label drift | Unified to `page-abstract-fallback` in all 4 mentions (Step 2.6 source list + Phase 3 report). |
| 🟡 4. arxiv row split | Now two rows: row 4 = `arxiv-api-abstract` (fresh API curl), row 5 = `page-abstract-fallback` (last-resort reuse of existing page abstract). Two distinct log labels. |
| ⚪ 5. resolver chain | Left as-is per your "not blocking, matches `/idea-creator` precedent" note. Happy to upgrade in a follow-up if you prefer. |

Ran a fresh-thread codex (gpt-5.5 xhigh, same approach as your review) over the updated file — it confirmed 1-4 RESOLVED and flagged a few additional internal-consistency items I also picked up in this push:
- `SOURCE_DEFAULT` summary matched the 5-row Phase 2.3 table
- `--force` exception list expanded to both protected sections, not just `Connections`
- Step 2.1 explicitly reads `graph/edges.jsonl` / `RESEARCH_BRIEF.md` / `CLAUDE.md` / `gap_map.md` (needed by the Claims and Relevance rules in Step 2.4)
- Removed nested backticks in Step 2.4 rule descriptions (Markdown was rendering weirdly)
- Dropped the MANIFEST.md mention since `wiki-enrich` isn't a defined stage in `shared-references/output-manifest.md` and adding one felt out of scope — `log.md` already carries the audit trail with provenance per paper

One thing the deeper codex pass noted that I'd like to flag transparently rather than expand scope unilaterally: `rebuild_query_pack` in `tools/research_wiki.py` currently only surfaces title/thesis + edges + gaps, **not** the `Reusable Ingredients` / `Limitations` sections that this skill fills. So calling `/idea-creator` after enrichment does see the new bodies *if it reads the paper pages directly* (which it can, via wiki paths), but `query_pack.md` alone wouldn't carry them. That's a `research_wiki.py` enhancement rather than a wiki-enrich one — happy to follow up with a separate PR if you'd like query_pack to start including `Reusable Ingredients` snippets.

Ready for another pass whenever you have time.

PR #246, opened 2026-05-23
feat: add manual-review MCP server for zero-cost cross-model review

ZBigFish:
## Motivation

ARIS's cross-model review requires both a Claude Code subscription (executor) AND a GPT Plus/Codex subscription (reviewer). This is expensive and creates a high barrier for users who only have one subscription. The reviewer workload is relatively light compared to the executor — it's a shame to lose cross-model review entirely just because of a missing API key.

## Solution

A new `manual-review` MCP server that acts as a human-in-the-loop bridge: when the pipeline needs cross-model review, it opens a local browser page where the user copies the prompt to **any** AI model (free ChatGPT, DeepSeek, Kimi, local models, etc.) and pastes the response back.

**Trade-off**: lose full automation, gain zero API cost and complete model flexibility.

## Features

| Feature | Details |
|---------|---------|
| Browser mode | Local web UI with one-click copy, bilingual (EN/CN), dark/light mode |
| File mode | For headless Linux/SSH: `prompt.md` → `response.md` file exchange |
| Thread continuity | `review_reply` shows previous exchanges for multi-round reviews |
| No timeout pressure | 24h default wait — designed for "research while sleeping" |
| Fixed port (17900) | User always knows where to find the page after accidental close |
| Non-blocking main loop | Interrupted calls don't block subsequent calls |
| Stability check | File mode requires content unchanged for 3s to prevent reading half-written files |
| Prompt fidelity | User sees the EXACT same prompt Codex would receive |
| Compatible return format | `{"threadId": "...", "content": "..."}` — matches Codex MCP exactly |
| Zero dependencies | Python stdlib only, matches `codex-image2` pattern |

## Usage

```bash
# Install
claude mcp add manual-review -s user -- python3 mcp-servers/manual-review/server.py

# Use with any reviewer-supporting skill
/auto-review-loop "topic" — reviewer: manual
/research-review "paper/" — reviewer: manual
```

After accidental page close, simply reopen: `http://127.0.0.1:17900`

## Verified Behavior

Tested in real Claude Code sessions:
- [x] Normal call → browser opens → submit → response returned correctly
- [x] Multiple consecutive calls work (port properly released between calls)
- [x] Interrupted call (Esc) → next call still works (non-blocking main loop + server_close)
- [x] Page left open after submit → next call unaffected
- [x] Return format `{"threadId", "content"}` matches Codex MCP exactly
- [x] 15 unit tests pass (protocol, HTTP, file mode, error handling, state management)
[...]

PR #267, opened 2026-05-28
Add gemini-review antigravity cli backend

ZGJY95:
Adds agy --log-file per call
Recovers response from Antigravity transcript when stdout times out
Supports AGY-generated text artifacts
Updates README


wanshuiyin:
Thanks @ZGJY95 — routing reviews through Gemini via the Antigravity CLI is a genuinely useful addition (gives Antigravity users a native cross-model reviewer without Codex MCP / Oracle). The CLI-spawning core is solid (`subprocess.run` with argv list + `shell=False`, UUID log paths under `STATE_DIR`).

But a fresh-thread cross-model review (codex gpt-5.5 xhigh) surfaced **2 blockers** + 2 smaller items around the transcript-recovery path. This is reviewer-backend code that other skills will trust to deliver verdicts, so the bar is the same we held the `manual-review` MCP (#246) to.

---

### 🔴 1. Transcript recovery trusts unvalidated conversation IDs + reads arbitrary local files

`latest_agy_conversation_id_for_cwd()` (server.py:285) reads `last_conversations.json` and `agy_transcript_paths()` (:314) builds a path from the conversation id. The id from the cache isn't validated, so a crafted/stale `last_conversations.json` entry could path-traverse out of `AGY_APP_DATA_DIR/brain/`. Worse: `extract_file_uri_paths()` reads any `file://...(.md/.txt/.json/.yaml)` artifact referenced in the transcript and returns it as the **reviewer verdict** — a stale or malicious transcript can fabricate a review result or exfiltrate arbitrary local text files into the response.

**Fix**: validate the conversation id against a strict `^[0-9A-Za-z_-]+$` allowlist (no `/`, `.`, `..`); confine artifact reads to within the resolved `AGY_APP_DATA_DIR/brain/<id>/` subtree (resolve + check `is_relative_to`); and bind the recovered transcript to *this* call (the log path / conversation id captured from this invocation's `--log-file`, not "latest for cwd") to avoid picking up an unrelated prior session.

### 🔴 2. No verification the backend is actually Gemini (cross-model invariant)

ARIS's core invariant: executor (Claude) ≠ reviewer family. This backend returns `"model": result.get("model")` / `"agy-cli"` (server.py:449) but the actual model `agy` talks to is whatever Antigravity is configured with — could silently be Claude or a GPT-OSS, which would make the "cross-model review" a same-family rubber-stamp. Please either (a) verify/assert the resolved model is a Gemini family id and warn/fail otherwise, or (b) clearly document that the user is responsible for configuring Antigravity to a non-executor family, and surface the actual model in the returned verdict so provenance is auditable.

### 🟡 3. `auto` backend now silently prefers `agy` when no API key

`resolve_backend()` (server.py:215) — the privacy/semantics of where prompts go changed (unpublished paper content can now flow to the Antigravity CLI + its transcript logs by default). This should be explicit opt-in, not the silent `auto` fallback.

### 🟡 4. Timeout doesn't kill the process group + PII retention

`subprocess.run(..., timeout=DEFAULT_TIMEOUT_SEC + 15)` (server.py:631) kills the direct child but not a process group, so an `agy`-spawned subprocess can zombie. Also the `--log-file` + transcript artifacts persist full prompts (which may contain unpublished paper text) with no documented retention / cleanup / permissions — please note this in the README and consider writing to a `0700` temp dir + cleanup on exit.

---

None of these are hard to fix — #1 and #2 are the must-do blockers (validate + confine the transcript path; assert/document the Gemini family). Happy to re-review once they're in. Codex thread: `019e71fa-8c69-7e93-9bba-cb64a26cc2cb`.

wanshuiyin:
**Addendum** — ran a deeper review pass (6 parallel dimension reviewers + a fresh independent cross-model audit). Confirms my earlier 4 items and surfaces a few more worth folding into the same revision. Ranked by confidence:

### Confirmed by two independent model families (highest confidence — please treat as must-fix)
- 🔴 **`file://` artifact arbitrary-read + verdict forgery** (server.py:333-340 → 367) — already in my first comment; reconfirmed.
- 🔴 **`conversation_id` path traversal** from `last_conversations.json` (server.py:300-316) — reconfirmed.
- 🔴 **cross-model invariant not enforced** (server.py:613-621, 666) — reconfirmed; also `:666` returns a hardcoded `"agy-cli"` placeholder model instead of the real one, so there's no provenance even when it IS Gemini.
- 🔴 **stale/raceable transcript recovery** ("latest for cwd" can pick up an unrelated prior session, server.py:381-383) — reconfirmed.
- 🔴 **timeout doesn't kill the process tree** (server.py:626) — `agy`/Gemini grandchildren leak; reconfirmed.
- 🔴 **prompt persisted to world-readable /tmp** (server.py:989, plus thread/job state at :481) — reconfirmed.

### New this pass — not in my first comment
- 🔴 **job/thread state path traversal** (server.py:~423) — `jobId` / `threadId` are concatenated into state-file paths without validation, so `../` enables reads/writes outside the state dir. This is a *separate* traversal surface from the `conversation_id` one — please validate both against `^[A-Za-z0-9_-]+$` + confine with `resolve()` + `is_relative_to()`.
- 🔴 **`communicate()` can block forever after a timeout** (server.py:626) — when the subprocess times out and is killed, if a grandchild still holds the stdout pipe open, the post-kill `communicate()` can hang indefinitely — defeating the timeout entirely. Kill the process *group* (`start_new_session=True` + `os.killpg`) and use a bounded `communicate(timeout=...)`.
- 🟡 **`review_reply` / `review_reply_start` schema omits required `threadId`** (server.py:~1062) — the tool schema requires only `prompt` but the implementation needs `threadId`; callers will hit a runtime error the schema doesn't predict.
- 🟡 **unbounded status `waitSeconds`** (server.py:~1165) — no max; a large value blocks this single-threaded MCP server.
- ⚪ **argv flag injection** (server.py:548-550, 613-621) — a `model` starting with `-` is passed positionally; no RCE (shell=False) but could be misread as a CLI flag. Cheap guard: reject leading `-`.

The CLI-spawning core is genuinely sound — the issues all cluster in the transcript-recovery + state-persistence + provenance layers. Once the 🔴 items are addressed I'm happy to re-review. (This pass: Claude parallel fan-out for breadth + Codex GPT-5.5 xhigh fresh independent audit for the cross-model jury — the two caught complementary blockers, which is itself why ARIS keeps the reviewer heterogeneous.)

ZGJY95:
Thanks for the detailed review. I pushed an update addressing the blocker set and the follow-up items.

Changes included:

- Hardened Antigravity transcript recovery:
  - conversation IDs are now allowlist-validated with `^[0-9A-Za-z_-]+$`
  - transcript/artifact reads are confined to `AGY_APP_DATA_DIR/brain/<conversation>/`
  - recovery is bound to the current invocation via the per-call `--log-file` plus an invocation nonce, so it no longer falls back to `last_conversations.json` / latest-for-cwd state

- Enforced reviewer provenance for `agy`:
  - the backend now recovers the actual Antigravity-selected model from the current log/transcript
  - it fails closed unless the recovered model is Gemini-family
  - it no longer returns an unauditable `"agy-cli"` placeholder
  - if a caller passes `model`, the response records it as `requested_model` with a warning, since current `agy` does not expose per-call model selection

- Tightened subprocess handling:
  - Gemini CLI and `agy` now run in a new process/session group
  - timeout cleanup kills the process tree and uses bounded post-timeout `communicate()`

- Hardened persisted state:
  - `jobId` / `threadId` now use the same safe-id validation and confined state paths
  - state/debug/agy log directories are created with owner-only permissions where possible
  - `agy` per-call logs are written under a private temp dir and cleaned up after the call
  - debug logging no longer writes full prompt payloads

- Fixed schema/runtime mismatches:
  - `review_reply` / `review_reply_start` schemas now require either `threadId` or `thread_id`
  - `review_status.waitSeconds` is capped in both schema and runtime
  - model overrides reject leading `-` and NUL-containing prompt/model values are rejected before subprocess launch

I also updated the README to document the explicit `agy` opt-in behavior, model provenance semantics, retention/privacy notes, and the fact that Antigravity model selection is external configuration rather than a per-call bridge option.

I also dogfooded the local `agy` path after the changes; it returned the actual recovered model and included `requested_model` + warning when a model was requested.

wanshuiyin (changes requested):
Thanks for the fast turnaround, @ZGJY95 — this is a strong revision. The subprocess hardening and the conversation-id confinement are exactly right; the allowlist + `resolve()`/`relative_to` containment is cleanly done.

I ran another fresh cross-model pass (codex gpt-5.5 xhigh) over the hardened `server.py`. Two of the three blockers are now closed; the most important one is still open, plus one new arbitrary-read vector. Requesting one more round before merge.

## ✅ Closed
- **Subprocess hang** — `Popen(shell=False)` + `communicate(timeout=)` + kill-process-group on timeout, returning an error rather than a success.
- **Direct conversation-dir traversal / arbitrary read** — `^[0-9A-Za-z_-]+$` `fullmatch` + `resolve()`-then-`relative_to`. 👍

## 🔴 Still must-fix — verdict forgery
The nonce binding is transcript-*wide*, so it doesn't actually scope recovery to *this* turn:
- `transcript_contains_nonce()` (`server.py:561-562`) is a whole-file substring check. Once any line contains this call's nonce, `extract_agy_response_from_transcript()` scans every `MODEL` line in the entire transcript (`633-660`) and can return a stale artifact written *before* the nonce.
- Recovered text can also override live CLI stdout when it's merely longer (`1045-1050`) — so a pre-existing long "looks great / ready to submit" artifact in the same conversation can replace the real verdict.
- The recovery helper is fail-open when `invocation_nonce is None` (`584-585`, `637-638`) — current call sites pass a nonce, but the primitive itself should refuse without one.

Why I'm gating hard here: the whole value of a cross-model review backend is that a "pass" is trustworthy. A path that returns a forged/stale "pass" defeats the cross-model invariant this backend exists to provide.

**Fix direction:**
- Scope recovery to the current turn: track the transcript line index where this invocation's nonce appears and only consider content **at/after** it — not the whole file.
- Prefer live stdout; use transcript recovery **only** when stdout is empty/timed-out — never let it override non-empty stdout.
- Make the nonce mandatory: if `invocation_nonce is None`, reject recovery (fail-closed), don't fall back to "latest conversation".

## 🟠 New — HIGH: `imagePaths` arbitrary local read / exfiltration
User-controlled image paths are `expanduser()`'d, `is_file()`-checked, then `read_bytes()`'d with no confinement (`372-383`). A symlink with an image extension lets the server read any readable file and ship the bytes to the Gemini API. Please confine image paths to an explicit workspace root (resolve + no-symlink open), or disable local image paths by default.

## 🟡 Smaller (fold into the same round)
- **No length cap** on `SAFE_ID_RE` (`68`) → path-length DoS via long `threadId`/`jobId`/conversation id. Add a sane max (e.g. 128).
- **Model name validated after invocation** — `selected_model` reaches `-m` (`905-909`) before the Gemini-family check (`939-942`). Validate before building argv.
- **Debug-log path can escape `STATE_DIR`** (`114-125`). Confine it under `STATE_DIR`.
- **Symlink TOCTOU** — `read_text_confined()` resolves+checks then opens separately (`257-264`). Open the fd first, then verify the fd's realpath (or use `O_NOFOLLOW`).

Really nice work overall — the architecture is right and you're one focused round away. Happy to re-review as soon as the forgery scoping + `imagePaths` confinement land. 🙏

ZGJY95:
Thanks again for the fresh pass. I ran an independent Codex gpt 5.5 review plus a Gemini 3.1 pro review over the latest hardened version.

The follow-up pass found no remaining blocker around the main transcript nonce scoping, but it did flag one worthwhile tightening: artifact fallback could still prefer a `file://` artifact referenced after the nonce. I pushed two additional commits to close that down:

- `699c5fa` — constrain artifact fallback
  - only parse artifact URIs from the final `PLANNER_RESPONSE`
  - prefer substantive inline planner text over artifact contents
  - stop collecting artifacts from intermediate/model/tool transcript lines

- `d896307` — bind artifacts to the invocation
  - when artifact fallback is needed, require the artifact mtime to be within the current `agy` invocation window
  - otherwise fail closed rather than returning a stale artifact

I also folded in two smaller hardening cleanups from the re-review:

- fd confinement now fails closed if the opened file descriptor cannot be re-verified; added `/dev/fd` / macOS `F_GETPATH` fallback in addition to `/proc/self/fd`
- API model path construction now handles `models/gemini-*` and `publishers/google/models/gemini-*` without double-prefixing `models/`

wanshuiyin:
Thanks again @ZGJY95 — this is a genuinely excellent revision, and thanks for sticking with it across several rounds. You've closed **6 of the 7 blockers**, and a few fixes go beyond what the review asked for:

- conversation-id and job/thread state paths are now allowlist-validated (`^[0-9A-Za-z_-]+$`) and confined with `resolve()` + `is_relative_to()`;
- the artifact reader uses `O_NOFOLLOW` + a regular-file (`S_ISREG`) check + a post-open fd-path recheck (TOCTOU/symlink-resistant) plus an mtime freshness gate — that's stronger than the original ask;
- `require_gemini_model` is now a hard fail (not a warning), and the **real** model is recovered from the log/transcript with an untrusted-text guard so prompt/response content can't spoof the model id;
- `auto` no longer silently routes to `agy`; the timeout now kills the process **group** with a bounded post-kill `communicate()`; job/thread state is written 0600 under a 0700 dir.

Really nice, careful work here. 🙏

A fresh cross-model pass (codex gpt-5.5 xhigh) confirms those six, and flags **one remaining blocker** + one hardening item:

### 🔴 1. Nonce binding can be satisfied by model-controlled content

The invocation nonce is appended to the **reviewer prompt** (`server.py:1091`), so the model sees it, and `transcript_lines_at_or_after_nonce` (`:652`) accepts the nonce on *any* JSONL line — including MODEL events. So the model can echo the nonce into its own output and satisfy the "transcript is bound to this invocation" check, which means the binding evidence rests on untrusted model output — exactly what shouldn't anchor a reviewer verdict.

This is mitigated by `conversation_id` already being scoped to this invocation's own `--log-file` (`:679`), so it's a binding-integrity gap rather than a clean arbitrary-transcript reopen — but for a reviewer backend the binding shouldn't rest on model-controlled text.

**Suggested fix:** parse the transcript as JSONL and accept the nonce only from **this invocation's user / bridge-written events** (match on the event whose role/source is the user request, not assistant/model output), instead of a substring scan over the whole transcript. You know the agy transcript schema best — if it helps, drop a small redacted sample of a *user* event line vs a *model* event line and we can sanity-check the predicate together.

### 🟡 2. Hardening: `load_private_env_file` mutates global `os.environ`

`load_private_env_file` (`:396`) imports **any** key from `~/.gemini/.env` into the process-wide `os.environ`, with no owner/mode/symlink check on the file. Suggest either allowlisting the credential keys you actually use (`GEMINI_API_KEY` / `GOOGLE_API_KEY`) or building a scoped subprocess `env` instead of mutating the global — and skipping the file if it's a symlink or group/world-writable.

Neither is hard, and #1 is the only must-fix. Happy to re-review as soon as it's in — thanks again for the thorough work on this backend.

Codex thread: `019e7bf8-d980-7ab0-847d-679261c16a7b`.

wanshuiyin:
Re-reviewed `646d39b3` ("Bind agy nonce to user events") — this closes the last blocker. 🎯

The nonce binding now parses each transcript line as JSON and accepts the nonce **only from a user/bridge-written event** (`transcript_step_is_user_event` gating `transcript_step_contains_nonce`), and I confirmed every other nonce path (`transcript_contains_nonce` → `agy_conversation_has_nonce` → `select_agy_conversation_id_from_state`) routes through that guard — so a model echoing the nonce in its own output can no longer satisfy the binding. A fresh cross-model pass (codex gpt-5.5 xhigh) independently agrees: **CLOSED** — under a trusted transcript schema, model output lands in assistant/model events and can't pass the user-event guard; the only residual is if Antigravity itself mislabeled a model turn as a user event, which is an upstream trust boundary outside the bridge's control (reasonable to rely on).

The 🟡 env-hardening item is in too — `PRIVATE_ENV_KEYS` allowlists only `GEMINI_API_KEY` / `GOOGLE_API_KEY`, with symlink / owner / mode / regular-file guards on `~/.gemini/.env`.

That clears everything across the review rounds (6 original blockers → the nonce binding → env hardening). Thanks for the thorough, responsive work on all of it — this is a genuinely solid, fail-closed cross-model reviewer backend now. **LGTM**, merging. 🙏

PR #288, opened 2026-06-07
fix(workflow): block incomplete idea-discovery runs

### LINKED ISSUE #285: idea-discovery: per-stage evidence gate — explicit BLOCKED instead of silent stage-skipping (follow-up to #284)
## Context

#284 reported `/idea-creator` skipping sub-skill calls and improvising inline. The direct causes (missing `Skill` grant + misleading Overview wording) were fixed in 07ca667. This issue tracks the remaining question from that report (Q4): **process enforcement** — workflows should not be able to silently degrade when a sub-skill stage is skipped.

## Problem

`/idea-discovery` (Workflow 1) chains

```
/research-lit → /idea-creator → /novelty-check → /research-review → /research-refine-pipeline
```

purely through natural-language instructions. Nothing structurally prevents the executor from skipping a stage and still emitting a complete-looking `IDEA_REPORT.md`. That is the same failure mode as #284, one level up — and **silent degradation is worse than a hard failure**, because the output looks done.

## Prior art (in-repo)

`/research-pipeline` already has the pattern: resumable runs record per-stage state to `.aris/runs/<run_id>.json`, splitting **`done`** (executor finished writing) from **`accepted`** (the stage's cross-model gate / deterministic verifier passed); resume re-validates any `done`-but-unaccepted stage. See [`shared-references/resumable-runs.md`](../blob/main/skills/shared-references/resumable-runs.md) and [`shared-references/acceptance-gate.md`](../blob/main/skills/shared-references/acceptance-gate.md) ("a loop can drive, never acquit").

## Proposal

1. **Per-stage evidence record**: after each sub-skill stage of `/idea-discovery`, record `{stage, status, artifact_path, trace_or_thread_id}` (artifacts: research-lit landscape, idea-creator candidates/report, novelty-check verdict, research-review output) — same `.aris/runs/` convention as `/research-pipeline`.
2. **Report gate**: the final `IDEA_REPORT.md` step MUST check the evidence records. Any missing stage ⇒ the report carries an explicit `BLOCKED: <stage> evidence missing` section and the run is **not** declared complete. Never silently proceed.
3. *(Optional hardening)* run [`tools/verify_papers.py`](../blob/main/tools/verify_papers.py) over the Landscape Summary citations, so a fabricated inline "survey" cannot pass as stage evidence.

## Acceptance criteria

- A skipped/failed sub-skill can no longer yield a complete-looking `IDEA_REPORT.md` — both the report and the run state say BLOCKED.
- Evidence records land in `.aris/runs/` (research-pipeline convention).
- `tools/check_skills_inventory.py` stays green; codex mirrors updated per convention.

Refs: #284 (Q4) · 07ca667

zerone0x:
## Summary

Fixes #285 by making Workflow 1 document an explicit per-stage evidence gate instead of allowing silent stage-skipping.

## Changes
- add `— resume <run_id>` + `RESUMABLE` guidance to `idea-discovery`
- define concrete `.aris/runs/<run_id>.json` phase/evidence mappings via `tools/run_state.py`
- require a visible `## BLOCKED` section in `IDEA_REPORT.md` whenever a stage is missing artifact/trace evidence
- update the Codex mirror in parallel, including composed-mode/output-hygiene guidance for folded sub-skill outputs

## Testing
- `python3 tools/check_skills_inventory.py`
- `git diff --check HEAD~1`

## Notes
- I also wrote `.claude/handoff.md` and did a local verification pass, but this runtime did not expose `sessions_spawn`, so I could not run the full isolated handoff-verify subagent flow here.

### DIFF
--- a/skills/idea-discovery/SKILL.md
+++ b/skills/idea-discovery/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: idea-discovery
 description: "Workflow 1: Full idea discovery pipeline to go from a broad research direction to validated, pilot-tested ideas. Use when user says \"找idea全流程\", \"idea discovery pipeline\", \"从零开始找方向\", or wants the complete idea exploration workflow."
-argument-hint: [research-direction]
+argument-hint: [research-direction] [— resume <run_id>]
 allowed-tools: Bash(*), Read, Write, Edit, Grep, Glob, WebSearch, WebFetch, Skill, mcp__codex__codex, mcp__codex__codex-reply
 ---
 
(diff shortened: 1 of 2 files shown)


wanshuiyin (changes requested):
Direction is exactly right — this implements #285 faithfully (per-stage evidence via `run_state.py`, explicit `## BLOCKED` instead of silent degradation, `done` vs `accepted` split per `resumable-runs.md`/`acceptance-gate.md`), and the composed-mode additions in the Codex mirror are genuine parity catch-up with mainline (verified `— composed:` is an established mainline convention in `research-review`/`idea-discovery`). Thank you — this will merge quickly once two deterministic test failures are fixed.

**Why you didn't see them:** `tools/check_skills_inventory.py` passes ✅, but the CI also runs `python -m pytest tests/test_codex_skill_mirror.py -q` (`.github/workflows/check-skills-inventory.yml`), which fails on this branch — and fork-PR workflows need maintainer approval to run, so there's no red ✗ visible yet. Reproduced locally on your head commit (bbb8771): **2 failed, 14 passed**.

### 1. `test_codex_skill_helper_commands_use_installed_aris_repo` — FAIL

`skills/skills-codex/idea-discovery/SKILL.md:60` hardcodes:

```bash
python3 tools/run_state.py start <root> <run_id> --phases ...
```

Codex-mirror skills must not assume helper scripts exist in the user's project (`python3 tools/` is a forbidden pattern in that test). The prose three lines above already states the correct resolver chain — the example just needs to use it:

```bash
RUN_STATE=""
for cand in .aris/tools/run_state.py tools/run_state.py "${ARIS_REPO:-}/tools/run_state.py"; do
  [ -f "$cand" ] && RUN_STATE="$cand" && break
done
[ -z "$RUN_STATE" ] && echo "WARN: run_state.py not resolved; continuing without run-state tracking" >&2
python3 "$RUN_STATE" start <root> <run_id> --phases "literature-survey,idea-generation,novelty-check,external-review,refine-plan,final-report"
```

### 2. `test_codex_shared_reference_links_exist` — FAIL

The mirror links `../../shared-references/resumable-runs.md` (×3) and `../../shared-references/output-composition.md` (×2), but neither file exists under `skills/skills-codex/shared-references/` — and the Codex install ships only the `skills-codex/` tree, so those links would dangle for Codex users (which is exactly what the test enforces).

Fix: copy both docs from `skills/shared-references/` into `skills/skills-codex/shared-references/`. I checked both — they contain no Codex-runtime-forbidden strings (`mcp__codex__codex` / `codex-reply` / `threadId`), so a verbatim copy is safe. Bonus: this also unblocks bringing `run_state.py` to the `research-pipeline` mirror later (it currently has no resumable-runs support at all).

### 3. Minor, non-blocking

Mainline `skills/idea-discovery/SKILL.md` also writes the literal `python3 tools/run_state.py start ...`. Mainline isn't test-gated on this, but the sibling convention (`skills/research-pipeline/SKILL.md:53-72`) states the resolver chain once and then refers to `run_state.py <subcommand>` abstractly. Suggest aligning for consistency.

After the two fixes, please run both locally:

```bash
python3 tools/check_skills_inventory.py
python3 -m pytest tests/test_codex_skill_mirror.py -q
```

PR #293, opened 2026-06-09
paper-plan: write outputs under .aris

deadpool66:
## Summary
- write paper-plan outputs under `.aris/outputs/paper-plan/`
- log paper-plan artifacts to `.aris/MANIFEST.md`
- avoid creating `PAPER_PLAN.md`, `PAPER_PLAN_*.md`, or `MANIFEST.md` in the project root unless explicitly requested
- document AAAI page-budget handling

## Notes
This keeps ARIS-generated artifacts out of the repository root by default.

### DIFF
--- a/skills/skills-codex/paper-plan/SKILL.md
+++ b/skills/skills-codex/paper-plan/SKILL.md
@@ -11,7 +11,7 @@ Generate a structured, section-by-section paper outline from: **$ARGUMENTS**
 
 - **REVIEWER_MODEL = `gpt-5.5`** — Model used via a secondary Codex agent for outline review. Must be an OpenAI model.
 - **TARGET_VENUE = `ICLR`** — Default venue. User can override (e.g., `/paper-plan "topic" — venue: NeurIPS`). Supported: `ICLR`, `NeurIPS`, `ICML`, `CVPR`, `ACL`, `AAAI`, `ACM`, `IEEE_JOURNAL` (IEEE Transactions / Letters), `IEEE_CONF` (IEEE conferences).
-- **MAX_PAGES** — Page limit. For ML conferences: main body to Conclusion end (excluding references, appendix). ICLR=9, NeurIPS=9, ICML=8. **For IEEE venues: references ARE included in page count.** IEEE journal Transactions ≈ 12-14 pages total, Letters ≈ 4-5 pages total; IEEE conference ≈ 5-8 pages total (including references).
+- **MAX_PAGES** — Page limit. For ML conferences: main body to Conclusion end (excluding references, appendix). ICLR=9, NeurIPS=9, ICML=8, AAAI=7 technical-content pages plus references unless the current AAAI CFP says otherwise. **For IEEE venues: references ARE included in page count.** IEEE journal Transactions ≈ 12-14 pages total, Letters ≈ 4-5 pages total; IEEE conference ≈ 5-8 pages total (including references).
 
 ## Inputs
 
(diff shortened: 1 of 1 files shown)


wanshuiyin (changes requested):
Thanks for this — the **motivation is legitimate** (ARIS-generated files cluttering the repo root is a real annoyance), and the **AAAI page-budget documentation is a clean, correct addition** I'd happily take on its own. But the output-relocation part can't merge as-is: it's a cross-skill *interface* change made in one mirror file, and it breaks the W3 pipeline for Codex users. Cross-checked the impact against the repo (and a second model, per ARIS's own cross-model discipline) — the read is consistent.

## Why this can't merge as-is

**1. It silently breaks ~5 downstream consumers (Codex pipeline).**
`PAPER_PLAN.md` is the fixed W3 contract. These skills read it and were **not** updated by this PR:
- `paper-write` — *"If no PAPER_PLAN.md exists, ask the user to run /paper-plan first"* (no `.aris/outputs/` fallback)
- `paper-writing` (11 refs), `paper-figure` (3 refs), `figure-spec` (1 ref), `paper-claim-audit`

A Codex user runs `/paper-plan` → plan lands in `.aris/outputs/paper-plan/PAPER_PLAN.md` → Codex `/paper-write` still looks for `PAPER_PLAN.md` and reports "no plan found." The producer moved; the consumers didn't.

**2. Mirror-only → Claude/Codex divergence.**
Only `skills/skills-codex/paper-plan/SKILL.md` changed. Mainline `skills/paper-plan/SKILL.md` still writes `PAPER_PLAN.md` to the project root. The mirror is supposed to track mainline behavior; this forks it.

**3. It invents a path convention that conflicts with the established one.**
`grep -rn ".aris/outputs/" skills/` → **zero hits anywhere**. The canonical layout in [`shared-references/output-versioning.md`](skills/shared-references/output-versioning.md) is root-level *stage* dirs (`idea-stage/`, `refine-logs/`, `review-stage/`, `paper/`). And [`output-manifest.md`](skills/shared-references/output-manifest.md) puts `MANIFEST.md` at **root**, created **only when a run produces >1 output** — so `.aris/MANIFEST.md` here is a second, contradictory manifest semantics.

## Constructive path forward

1. **Split this PR**: keep only the AAAI page-budget edit (clean, mergeable now); drop the `.aris/outputs/` + `.aris/MANIFEST.md` changes.
2. **Open an RFC/issue for a repo-wide output layout** before moving any paths. The `.aris/outputs/<stage>/` idea is genuinely worth considering — consolidating under `.aris/` is arguably cleaner than scattering stage dirs in root — but it's an output-*protocol* decision: where do outputs live, where does the manifest live, which artifacts are user-editable first-class files, and how do we fall back to legacy root/stage paths.
3. **If the RFC lands, do one repo-wide migration PR**: update `output-versioning.md`/`output-manifest.md` first, then move producers *and* consumers together (paper-plan, paper-write, paper-writing, paper-figure, figure-spec, paper-claim-audit), mainline **and** Codex mirrors, plus docs/templates, with a backward-compatible read fallback for existing `PAPER_PLAN.md` files.

Happy to fast-track the AAAI-only version. The layout change I'd love to see land too — just as a coordinated protocol change, not a single mirror edit.

deadpool66:
Thanks for the careful review. I’ve updated this PR to keep only the AAAI page-budget documentation change and removed the .aris/outputs/ / manifest relocation changes from the PR.

The output-layout change makes sense as a broader protocol/RFC item, so I’ll keep that out of this PR.

wanshuiyin (approved):
Approving — thanks for the fast, gracious turnaround. The PR is now exactly the clean subset suggested: the AAAI page-budget documentation (correct as stated — AAAI main track is 7 technical-content pages plus references, and the "unless the current AAAI CFP says otherwise" hedge is the right call since these limits shift year to year), with the `.aris/outputs/` relocation cleanly withdrawn pending an RFC. Verified on the PR head: mirror tests 16/16, skills inventory consistent.

One housekeeping note: since this PR touches only the Codex mirror, I'll follow up directly on main with the identical two-line AAAI edit to mainline `skills/paper-plan/SKILL.md` so the mirror doesn't carry more venue info than the source of truth.

And the output-layout RFC remains genuinely welcome if you want to drive it — consolidating generated artifacts under `.aris/` is a direction worth designing properly.

PR #294, opened 2026-06-13
feat: add factual_audit.sh - automated truth check before review

hxm2023:
An 8-check automated audit that verifies claims before sending to cross-model review. Catches provable errors deterministically.

Checks: reproduce.sh, >=3 checkpoints, seed consistency, test set size, epoch count, baseline sources, paper figures, evidence existence (numbers in paper found in experiment results).

Exit 0 = clean. Exit 1 = issues found.

### DIFF
--- a/tools/factual_audit.sh
+++ b/tools/factual_audit.sh
@@ -1,0 +1,116 @@
+#!/usr/bin/env bash
+# ARIS Factual Audit — automated truth check before Codex MCP review.
+# Usage: bash factual_audit.sh <paper_dir> <experiment_log_dir>
+# Exit 0 = all checks passed. Exit 1 = issues found (sent to Codex as context).
+set -euo pipefail
+
+PAPER_DIR="${1:-paper}"
+EXPERIMENT_DIR="${2:-deep-experiment-logs}"
+ISSUES=0
+
+RED='\033[0;31m'; GREEN='\033[0;32m'; NC='\033[0m'
+pass() { echo -e "${GREEN}[AUDIT OK]${NC} $*"; }
+fail() { echo -e "${RED}[AUDIT FAIL]${NC} $*"; ISSUES=$((ISSUES + 1)); }
+
+echo "══════════════════════════════════════════"
+echo "  FACTUAL AUDIT"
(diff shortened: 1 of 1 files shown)


wanshuiyin (changes requested):
Thanks for this — the instinct behind it is good. A cheap, deterministic **preflight gate that catches provable problems before spending a cross-model review** is genuinely aligned with where ARIS is heading (see the per-stage evidence gate in #288 and the `done` vs `accepted` split in `run_state.py`).

The blocker is that the script hardcodes one specific project's layout and thresholds rather than ARIS's conventions, so on a real ARIS project it would fail almost every check:

- **Paths that aren't ARIS conventions:** `reproduce.sh`, `checkpoints/*.pt`, `deep-experiment-logs/ROUND_*/results/metrics.json`, `BASELINE_SOURCES.md`. ARIS's documented output layout (`shared-references/output-versioning.md`) is `idea-stage/`, `refine-logs/`, `review-stage/`, `paper/` — none of these files exist in it.
- **Magic thresholds baked in as universal truth:** `>=3 checkpoints`, `>=5 seeds`, `>=200 test`, `>=100 epochs`, `>=4 figures`. Those are reasonable for some projects but are domain/venue-specific value judgments, not facts.
- **Check 8** greps every decimal in the paper (`grep -oh '[0-9]+\.[0-9]+'`) against the results dir — in practice that false-positives on p-values, version numbers, hyperparameters, etc.

There's also overlap with what ARIS already does semantically: `experiment-audit` (fake ground truth / normalization fraud / phantom results) and `paper-claim-audit` (zero-context check that paper numbers match results). ARIS deliberately routes correctness judgments through a cross-model reviewer rather than hardcoded bash thresholds, precisely to avoid brittle false pass/fail.

If you'd like to pursue the preflight-gate idea, the version that would fit is: **contract-aware, configurable, and non-blocking** — verify only ARIS's real output contracts (does the stage's expected artifact exist?), keep thresholds in a config the user can set, and leave the quality verdict to the semantic audit skills. Happy to help scope that if you want to take it on. Thanks for the energy here!

PR #295, opened 2026-06-13
feat: add run_state.sh - resume pipeline after interruption

hxm2023:
Tracks Phase completion state (done/accepted/skipped) in a simple JSON file. After each Phase completes, write state. On restart, read state to know where to resume.

Usage:
  bash tools/run_state.sh set <phase> done
  bash tools/run_state.sh accept <phase> <verdict_id> <reviewer>
  bash tools/run_state.sh get
  bash tools/run_state.sh resume   # prints next phase or 'done'

Solves the pain point of long pipeline runs that crash mid-way with no record of which Phases already finished.

Signed-off-by: hxm2023

### DIFF
--- a/tools/run_state.sh
+++ b/tools/run_state.sh
@@ -1,0 +1,92 @@
+#!/usr/bin/env bash
+# ARIS Run State — track pipeline phase completion for resume.
+# Usage: bash run_state.sh set <phase> done         (phase completed by executor)
+#        bash run_state.sh accept <phase> <verdict_id> <reviewer>  (cross-model passed)
+#        bash run_state.sh get                       (print current state)
+#        bash run_state.sh resume                    (print next phase to run, or "done")
+set -euo pipefail
+
+STATE_DIR=".aris/runs"
+STATE_FILE="$STATE_DIR/run_state.json"
+mkdir -p "$STATE_DIR"
+
+# Initialize if missing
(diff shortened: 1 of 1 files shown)


wanshuiyin (changes requested):
Thanks for this, and for tackling the resumable-pipeline problem — it's a real pain point and your `set` / `accept` / `get` / `resume` interface is exactly the right shape.

One thing to flag: ARIS already ships this as **`tools/run_state.py`** (same `start`/`set`/`accept`/`get`/`resume` semantics, same `done` vs `accepted` + `verdict_id` + `reviewer` split). It's already wired into `research-pipeline`, `shared-references/integration-contract.md`, and `shared-references/resumable-runs.md`, and PR #288 is extending it into `idea-discovery`. So a second, independent state-machine in bash would fork that contract — and since `run_state.sh` reads/writes the JSON via inline `python3` heredocs anyway, it isn't python-free in practice.

**There's an easy way to land this, though:** turn `run_state.sh` into a thin shell entry point that delegates to the canonical implementation, e.g.

```bash
#!/usr/bin/env bash
exec python3 "$(dirname "$0")/run_state.py" "$@"
```

That gives shell-first / Codex-CLI users the `.sh` entry point they'd reach for, with zero logic duplication and one source of truth. If you reshape the PR that way (and note in the description that `run_state.py` stays canonical), I'm happy to merge it. Thanks again!

PR #296, opened 2026-06-13
feat: add gate_check.sh — pipeline phase verification tool

hxm2023:
A 60+ checkpoint binary audit that verifies every Phase's outputs exist on disk before the pipeline proceeds. Runs entirely in bash with zero dependencies.

Usage:
  bash tools/gate_check.sh 0-4, 2-round, all

Exit 0 = PASS, proceed. Exit 1 = FAIL, fix before proceeding.

Checks: KB files, IDEA_REPORT, per-round 6 files, bibliography, figures, paper output, citations, template, review stage.

Auto-logs via tee.

Signed-off-by: hxm2023

### DIFF
--- a/tools/gate_check.sh
+++ b/tools/gate_check.sh
@@ -1,0 +1,293 @@
+#!/usr/bin/env bash
+# ARIS Gate Check — verify Phase outputs before pipeline transitions.
+# Usage: bash gate_check.sh <phase>
+#   phase 0: check knowledge base outputs
+#   phase 1: check idea discovery outputs
+#   phase 2: check deep experiment outputs (per-round + final)
+#   phase 2-round: check current round's 6 mandatory files
+#   phase 3: check paper writing outputs
+#   phase 4: check review outputs
+# Exit 0 = gate PASS, proceed.  Exit 1 = gate FAIL, go back.
+set -euo pipefail
+
+# Auto-log to deep-experiment-logs/ or project root
+LOG_DIR="${ARIS_LOG_DIR:-deep-experiment-logs}"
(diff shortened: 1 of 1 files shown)


wanshuiyin (changes requested):
Thanks for this — it's the same good "verify outputs exist before the pipeline proceeds" instinct as #294, and a binary, zero-dependency gate is an appealing idea in principle.

Same core issue, though: the checks are written against a pipeline structure that isn't ARIS's, so they'd fail on a real ARIS project. A couple of concrete examples:

- Phase 0 checks `research-wiki/knowledge_base/domain_overview.md`, `metrics_and_baselines.md`, `field_conventions.md`, `index.json`, `search.py`. ARIS's `research-wiki` isn't structured that way — it's `papers/`, `ideas/`, `experiments/`, `claims/`, `edges/`. There's no `knowledge_base/` with those files.
- The experiment phases assume `deep-experiment-logs/ROUND_*/` (23 references), which isn't an ARIS output convention either (ARIS uses `refine-logs/`, `review-stage/`, `paper/`).

So this reads as a gate tailored to your own project's directory conventions rather than to ARIS's documented contracts in `shared-references/output-versioning.md`.

As with #294, the idea is worth pursuing if it's reframed to ARIS's real contracts — a contract-aware, configurable, non-blocking preflight that checks for the artifacts ARIS actually produces, and defers quality/correctness to the cross-model audit skills (`experiment-audit`, `paper-claim-audit`, `citation-audit`). I'd suggest consolidating #294 and #296 into a single such tool rather than two overlapping gates. Thanks again for the effort!

PR #298, opened 2026-06-18
fix(render-html): prevent MathJax formula truncation on bare '<' in math

Dennis-Huangm:
## Problem

In the `render-html` skill, a bare `<` immediately followed by a letter inside a math body (e.g. `y_{<t}`, `\cos(...) < 0`) is parsed by the browser as an HTML start tag. It silently swallows the rest of the formula — including the closing `$$` delimiter — so MathJax never receives a complete delimiter pair and renders the raw LaTeX, visibly **truncated at the `<`**.

This shows up on common notation like `y_{<t}` (token-position conditioning), `> \tau` thresholds, and `< 0` sign conditions inside `cases`/`align` environments.

## Fix

Rewrite `<` / `>` to `\lt ` / `\gt ` at the **LaTeX level** inside math bodies before stashing. This removes the `<` character entirely, so no parser can ever read it as a tag, and it renders identically.

- Trailing space prevents macro-name merging (`<t` → `\lt t`, not `\ltt`).
- `&` is deliberately left untouched — it is the alignment character in `cases`/`align` and must pass through verbatim.
- Chosen over HTML-entity escaping (`&lt;`/`&gt;`), which relies on the browser decoding the entity back into `textContent` before MathJax reads it and is fragile across caches / decode paths.

Applied to both copies of the script:
- `skills/render-html/scripts/render_html.py`
- `skills/skills-codex/render-html/scripts/render_html.py`

## Test

- Both files pass `python -m py_compile`.
- Verified against a document using `y_{<t}`, `> \tau`, and a `cases` env with `\cos(...) < 0`: formulas now render fully with `&` alignment intact and zero literal `<` surviving into the HTML math.

### DIFF
--- a/skills/render-html/scripts/render_html.py
+++ b/skills/render-html/scripts/render_html.py
@@ -147,16 +147,10 @@ def render_inline(text: str) -> str:
 
     text = _RE_CODE_INLINE.sub(_code_sub, text)
 
-    # HTML-escape <, >, & inside math bodies. Before MathJax runs, a bare "<t"
-    # (e.g. y_{<t}) is parsed by the browser as an HTML start tag and silently
-    # swallows the rest of the formula (including the closing $$), so MathJax
-    # never sees a complete delimiter pair and dumps the source as raw text.
-    # MathJax v3 reads the *decoded* DOM text node (HTMLDomStrings.handleText ->
-    # adaptor value() -> node.nodeValue), so the browser turns &lt;/&gt;/&amp;
-    # back into </>/& before MathJax sees them — the TeX token stream is
-    # byte-identical and rendering is unchanged. This is escape-transparent and,
-    # unlike a TeX-level \lt/\gt rewrite, leaves literal "<" in \text{}/\texttt{}
-    # and the cases/align alignment "&" intact (quote=False keeps quotes raw).
+    # HTML-escape <, >, & inside math bodies: a bare "<t" (e.g. y_{<t}) is
+    # otherwise parsed as an HTML start tag and eats the rest of the formula
+    # before MathJax runs. MathJax v3 reads the decoded DOM text, so the escape
+    # is transparent to the TeX it sees (and & for cases/align stays intact).
     # 1b. Display math (passthrough; MathJax will render).
     def _md_sub(m: re.Match[str]) -> str:
         body = html_lib.escape(m.group(1), quote=False)
--- a/skills/render-html/scripts/render_html.py
+++ b/skills/render-html/scripts/render_html.py
@@ -147,16 +147,30 @@ def render_inline(text: str) -> str:
 
     text = _RE_CODE_INLINE.sub(_code_sub, text)
 
+    # Neutralize < and > inside math bodies at the LaTeX level. Before MathJax
+    # runs, a bare "<t" (e.g. y_{<t}) is parsed by the browser as an HTML start
+    # tag and silently swallows the rest of the formula (including the closing
+    # $$), so MathJax never sees a complete delimiter pair and dumps the source
+    # as raw text. HTML-entity escaping (&lt;/&gt;) relies on the browser
+    # decoding the entity back into textContent before MathJax reads it, which
+    # is fragile across caches/decode paths. Rewriting to \lt / \gt removes the
+    # "<" character entirely — no character a parser could ever read as a tag —
+    # and renders identically. Trailing space prevents macro-name merging
+    # (e.g. "<t" -> "\lt t", not "\ltt"). Do NOT touch & — it is the alignment
+    # char in cases/align environments and must pass through verbatim.
+    def _math_safe(body: str) -> str:
+        return body.replace("<", "\\lt ").replace(">", "\\gt ")
+
     # 1b. Display math (passthrough; MathJax will render).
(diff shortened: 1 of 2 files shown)


wanshuiyin (changes requested):
Thank you — this is a **real, live bug** and a sharp catch. A bare `<t` inside math really is eaten by the browser's HTML tokenizer before MathJax ever runs, and `main` currently has no protection at all (`body = m.group(1)`). It also correctly touches **both** copies (mainline + Codex mirror), which is exactly right. The only thing I'd change is the *layer* the fix operates at.

I'd recommend solving it with **HTML-entity escaping** rather than a TeX-level `\lt`/`\gt` rewrite. I cross-checked this with a second model (GPT-5.5, the repo's own cross-model review path) against the MathJax v3 source, and the conclusion was clear.

### Why escape at the HTML layer

```python
# 1b. Display math (passthrough; MathJax will render). HTML-escape <, >, &
# inside the body so the browser does not parse e.g. "y_{<t}" as an HTML tag
# (which silently eats the rest of the formula). MathJax reads the decoded
# text content, so the LaTeX it sees is unchanged — this is escape-transparent.
def _md_sub(m: re.Match[str]) -> str:
    body = html_lib.escape(m.group(1), quote=False)
    return store(f"$${body}$$")

# 1c. Inline math — same escaping.
def _mi_sub(m: re.Match[str]) -> str:
    body = html_lib.escape(m.group(1), quote=False)
    return store(f"${body}$")
```

1. **MathJax v3 reads decoded DOM text, so escaping is transparent.** Its TeX DOM flow (`HTMLDocument.findMathFromStrings` → `HTMLDomStrings.handleText` → the adaptor's `value()` → `node.nodeValue`) reads the **parsed** text node, not the raw HTML string. `&lt;` in a text context is decoded to `<` by the browser *before* MathJax sees it, so MathJax receives the original TeX character. The same round-trip makes `&amp;` → `&` work, so this approach handles the `cases`/`align` alignment `&` correctly too — no need to special-case it.

2. **The "fragile across caches/decode paths" concern doesn't hold here.** Character-reference decoding in an HTML text context is deterministic per the WHATWG tokenizer (data state → character-reference state), not a heuristic. Caching the final HTML doesn't change the browser's parse. (If some layer re-injected already-decoded text as `innerHTML`, that would be a double-injection bug in that layer, not a property of entity escaping.)

3. **`\lt`/`\gt` rewriting moves an HTML problem into the TeX source and adds edge cases.** A blanket `.replace("<", "\\lt ")` is fine for plain relations (`y_{<t}`, `\cos(x) < 0`), but it rewrites the TeX language everywhere: inside `\text{a < b}`, `\texttt{<tag>}`, or `\operatorname{a<b}` the `<` is literal text and `\lt` is not a valid text-mode synonym; `<<` becomes two relation macros (not the original two glyphs, and not `\ll`); `->`/`<-` get the `>`/`<` macro-ized. Leaving `&` untouched also still exposes it to HTML character-reference parsing. Escaping at the HTML layer avoids all of this because the TeX token stream is byte-identical after the browser decodes it.

4. **Bonus: it re-syncs the helper.** The sibling [ARIS-in-AI-Offer](https://github.com/wanshuiyin/ARIS-in-AI-Offer) copy of `render_html.py` already uses exactly this escape-based fix, so adopting it here keeps the canonical helper consistent across repos.

If you switch both `_md_sub` and `_mi_sub` to `html_lib.escape(m.group(1), quote=False)` (mainline **and** the Codex mirror), I'm happy to merge right away. Thanks again for finding and fixing the truncation — the diagnosis is spot on, it's just a matter of doing it one layer up.

Dennis-Huangm:
Thanks for the thorough cross-model review — fully convinced. Pushed `28bc6e9`: both `_md_sub` and `_mi_sub` now use `html_lib.escape(m.group(1), quote=False)` in the mainline copy **and** the Codex mirror, dropping the `_math_safe` \lt/\gt rewrite entirely. The comment block is rewritten to explain the MathJax-v3 decoded-DOM-text rationale and to note that `&` (cases/align alignment) and literal `<` in `\text{}`/\texttt{}` are now preserved. Both files `py_compile` clean. This also matches the escape-based fix already in ARIS-in-AI-Offer. Ready for another look.

wanshuiyin (approved):
Perfect — thank you for the fast, gracious turnaround. `28bc6e9` is exactly right: both `_md_sub` and `_mi_sub` now escape at the HTML layer in the mainline copy **and** the Codex mirror, and `_math_safe` is gone.

Verified on the PR head:
- `html_lib.escape(m.group(1), quote=False)` in both functions, both files; no `\lt`/`\gt` rewrite remaining.
- `py_compile` clean; `tests/test_codex_skill_mirror.py` 16/16; `check_skills_inventory.py` green.
- Behavior spot-check: `y_{<t}` → `y_{&lt;t}`, `\cos(x) < 0` → `\cos(x) &lt; 0`, `a &= b` → `a &amp;= b` (alignment `&` round-trips), and crucially `\text{if a<b}` → `\text{if a&lt;b}` — the browser decodes each back to the original TeX before MathJax reads it, so even literal `<` inside `\text{}` is preserved. That's the case the `\lt` rewrite would have broken.

Merging. Nice catch on a real truncation bug.

PR #327, opened 2026-06-30
Fix claude-review subprocess decoding on Windows

YQYANG2233:
## Summary
Fixes claude-review crashes on Windows when Claude CLI emits UTF-8 output that Python tries to decode with the system default GBK codec.

Changes:
- Explicitly decode Claude CLI subprocess output as UTF-8 with errors="replace".
- Treat empty stdout as "" before JSON parsing.
- Guard stderr.strip() with (result.stderr or "").strip() to avoid secondary NoneType crashes.

## Motivation
On Windows, background review_start jobs could fail with: UnicodeDecodeError: 'gbk' codec can't decode byte ...
That decode failure left result.stderr as None, and the MCP server then crashed again with: 'NoneType' object has no attribute 'strip'
This made the exposed error misleading and caused review jobs to fail even when the issue was only subprocess output decoding.

## Validation
- python -m py_compile server.py
- Verified review_start + review_status end to end with a minimal prompt.
- Confirmed the fix works in practice on Windows.

### DIFF
--- a/mcp-servers/claude-review/server.py
+++ b/mcp-servers/claude-review/server.py
@@ -287,6 +287,8 @@ def run_claude_review(
             cmd,
             capture_output=True,
             text=True,
+            encoding='utf-8',
+            errors='replace',
             stdin=subprocess.DEVNULL,
             timeout=DEFAULT_TIMEOUT_SEC,
             check=False,
@@ -294,9 +296,9 @@ def run_claude_review(
     except subprocess.TimeoutExpired:
         return None, f"Claude review timed out after {DEFAULT_TIMEOUT_SEC} seconds"
 
-    payload, parse_error = parse_claude_json(result.stdout)
+    payload, parse_error = parse_claude_json(result.stdout or "")
     if parse_error:
-        stderr = result.stderr.strip()
+        stderr = (result.stderr or "").strip()
         message = parse_error if not stderr else f"{parse_error}. stderr: {stderr}"
         return None, message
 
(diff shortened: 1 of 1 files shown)


wanshuiyin (approved):
Thanks @YQYANG2233 — clean, correct fix. 🙏

The root cause is right: with `text=True` and no `encoding`, `subprocess.run` decodes the Claude CLI's UTF-8 output via the platform default codec (GBK on Windows) → `UnicodeDecodeError`, and the follow-on `result.stderr.strip()` then hits `None`. Pinning `encoding='utf-8'` (+ `errors='replace'`) and the `(result.stderr or "")` guards resolve both.

I checked the fix is **complete**, not just one spot: `run_claude_review` is the only place Claude CLI output is decoded in-process, and it's the path taken by both the synchronous tools **and** the background `--run-job` worker that `review_start` spawns — so the background path is covered. The `subprocess.Popen` worker only redirects the child's stdout/stderr to `.log` files (raw fd redirect, no Python text decode), and those logs are never read back, so there's no second GBK site.

Reviewed with a second model (cross-model check) — same verdict: approve.

One **non-blocking** note: `errors='replace'` is a sensible best-effort choice (return the review text rather than crash). If you'd ever rather *surface* a genuinely corrupt payload than silently replace bytes, you could drop it (strict) and `except UnicodeDecodeError` into a clear error — but that's optional and not required for this fix. A one-line comment noting `errors='replace'` is intentional would be a nice touch.

LGTM 🚀

PR #360, opened 2026-07-12
fix: use native Copilot cross-model reviewer by default

### LINKED ISSUE #258: 【feat】支持使用 Copilot CLI 内置 subagent 进行审阅，减少对 MCP 的硬依赖
Background

  当前多个 skill 的审阅流程依赖外部 mcp__codex__chat / mcp__llm-chat__chat 来完成 reviewer 调用。
  但在 Copilot CLI 环境下，其实已经可以直接使用内置 subagent，并指定 gpt-5.4 完成审阅任务，无需额外走 MCP。

  Problem

  现状带来几个问题：

   1. 额外依赖：用户必须先配置 mcp，否则审阅链路不可用。
   2. 重复路由：明明已经运行在 Copilot CLI 内部，却还要把审阅再转发到外部 MCP。
   3. 上下文传递低效：很多本地上下文只能拼进 prompt，而不是直接交给 subagent 使用。
   4. 稳定性更差：MCP 配置、网络、模型路由任一环节出问题，都会阻断 review 流程。

  Expected Behavior

   - 在 Copilot CLI 中运行 skill 时，如果用户没有显式指定 reviewer backend： - 默认使用 Copilot CLI 内置 subagent + GPT-5.4
   - 如果用户明确要求外部 reviewer： - 仍可切换到 mcp__codex__chat

Functionhx:
## Summary

Closes #258.

Makes `/auto-review-loop` use Copilot CLI's built-in complementary
`rubber-duck` subagent by default when no reviewer override is supplied. The
reviewer model is selected by Copilot rather than fixed to GPT-5.4; ARIS accepts
the result only when persisted host events prove that the executor and reviewer
belong to known, different model families.

Explicit `codex`, `oracle-pro`, `agy`, and `manual` choices still bypass native
activation. Explicit `copilot` retains the existing custom-agent compatibility
drive mode and its external finalizer.

## What changed

- Add a two-call marker/challenge protocol that binds the current project and
  Copilot root session without relying on the unshipped `COPILOT_CLI` variable.
- Dispatch the real native `task(agent_type=rubber-duck)` tool; slash prompts,
  top-level `copilot --agent rubber-duck`, nested calls, stale sessions, and
  generic-agent lookalikes are rejected.
- Add load-bearing evidence that revalidates the marker completion, native
  subagent lifecycle, actual host-reported model IDs, event order/prefix, and
  raw tool response.
- Wire verified native responses directly into the executable stop gate and
  trace schema. Positive native verdicts need no MCP/manual finalizer; missing,
  malformed, same-family, or unknown-family evidence fails closed.
- Revalidate the host-bound executor model before any external fallback and
  route the fallback through the stricter cross-family external-acquittal gate.
- Use unique `run_id + round + purpose` evidence paths so activation is not
  duplicated and prior audit artifacts are never overwritten.
- Keep failed pre-evidence native dispatches traceable without granting them
  model provenance or stop authority.
- Preserve default Codex behavior outside a bound Copilot session and keep
  native helper loading lazy for explicit non-native backends.
- Update routing, tracing, adaptation, integration-contract, README, and agent
  documentation.

## Validation

- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest -q`
  - `600 passed, 20 skipped, 4 subtests passed`
- Native/gate/installer focused suite: `92 passed`
- `python3 -m py_compile tools/copilot_native_evidence.py tools/review_gate.py`
- `bash -n tools/save_trace.sh`
- `python3 tools/check_skills_inventory.py`
- `git diff --check`

Live Copilot CLI 1.0.70 validation bound a real root session, captured its
host-reported `gpt-5-mini` executor model, and emitted the real native
`task(agent_type=rubber-duck)` invocation. The current account/session did not
expose that built-in agent, so Copilot returned `Unknown agent_type:
rubber-duck`; ARIS correctly produced no evidence/response and failed closed.
The successful cross-family lifecycle, response extraction, gate, trace, and
tamper cases are covered with session-event fixtures matching Copilot's
persisted schema.


wanshuiyin (changes requested):
Thank you for this PR — and for clearly having read the house contracts before writing it. The fail-closed `REVIEW_UNAVAILABLE` posture, the cross-family router with an explicit family-derivation table, the injection-safe env-passing in `save_trace.sh`, the honest "Verified vs Assumed" table, and scoping to a vertical slice of `/auto-review-loop` are all exactly the right instincts. #258 is a real need and we want this direction to land.

That said, we verified the integration contract against the primary Copilot CLI sources, and ran an independent cross-model review of the diff. There are two layers of issues that need author-side rework before this can merge, so I'm requesting changes rather than patching it ourselves — the fixes need access to a live Copilot CLI environment, which you have and we don't.

## Layer 1 — the integration contract doesn't match Copilot CLI as it exists today

1. **`COPILOT_CLI` env var does not exist.** It is a *proposal* in [github/copilot-cli#2107](https://github.com/github/copilot-cli/issues/2107) (still open). The [documented env vars](https://deepwiki.com/github/copilot-cli/5.2-environment-variables) are all configuration the CLI *reads* (`GH_TOKEN`, `HTTPS_PROXY`, …); per #2107 the only incidental runtime signals are `COPILOT_CLI_BINARY_VERSION` (SEA builds only) and `COPILOT_RUN_APP` (internal loader detail) — neither stable. As written, auto-detection never triggers and the headline feature is silently inert (it falls back to codex, which is at least fail-safe). Suggested fix: drop env-sniffing for now — make `--reviewer: copilot` explicit-only, and leave a note to revisit auto-detection when #2107 lands (or cite a real, stable signal with a source).

2. **Custom agent files must use the `.agent.md` extension.** The [official CLI docs](https://docs.github.com/en/copilot/how-tos/copilot-cli/customize-copilot/create-custom-agents-for-cli) state: "Each custom agent is defined by a Markdown file with an `.agent.md` extension." The PR ships `aris-reviewer-openai.md` / `aris-reviewer-claude.md`, which the CLI will not discover. (The directories are right: `.github/agents/` project-level, `~/.copilot/agents/` user-level.) One-line rename — plus the installer manifest entries.

3. **The `task: agent_type: "<profile>"` contract is marked "Verified" but isn't documented.** The [documented invocation mechanisms](https://docs.github.com/en/copilot/how-tos/copilot-cli/use-copilot-cli/invoke-custom-agents) are: `/agent`, natural-language instruction ("Use the security-auditor agent on…"), automatic inference, and programmatic `copilot --agent NAME --prompt "..."`. The `task` tool is documented only as general-purpose subtask execution; its parameter schema is not public. Your shape may happen to work — but please either (a) test it in a live session and paste the evidence, or (b) switch to the documented programmatic form (`copilot --agent aris-reviewer-openai --prompt "..."` as a subprocess — structurally the same as our `codex exec` pattern, and arguably cleaner). Either way, downgrade the claim from "Verified" until there's evidence.

## Layer 2 — reviewer-doctrine gaps (from an independent cross-model review of the diff)

1. **The review floor is bypassed.** The copilot profiles pin `gpt-5.4` / `claude-sonnet-4.5` with no reasoning-effort control ("depends on profile defaults"), yet this backend can issue the verdict that terminates `/auto-review-loop`. Our routing contract sets an `xhigh` floor for verdict-bearing reviews and allows `gpt-5.4` only as an explicit legacy override. If effort genuinely cannot be pinned through profiles, the backend needs to be honest about the degradation — e.g. verdicts from it are recorded as effort-unpinned and the loop's acceptance is marked accordingly, or the backend is documented as drive-only (it can iterate, a codex/manual backend must acquit).

2. **Cross-family verification is self-attested.** `--executor-model` is caller-declared and `independence_verified: true` is trusted as passed (`save_trace.sh` records whatever it's given). At minimum: parse the selected profile's frontmatter and record its actual pinned model as `requested_reviewer_model` (the routing doc promises this field but `save_trace.sh` has no parameter for it), record `reported_reviewer_model` when the tool surfaces it, and never write `independence_verified: true` unless derived from those — otherwise record `"unverified"`.

3. **`REVIEWER_MEMORY.md` breaks reviewer-owned continuity.** The executor appends/curates the file the reviewer "remembers" — the party under review is editing the reviewer's memory (and the default `medium` difficulty skips memory updates entirely, so round 2+ loses prior concerns). Codex-backend continuity is server-side and executor-untouchable. Suggested shape: the reviewer's full raw response is appended verbatim (never summarized by the executor), the file is append-only, and each round's trace records the file's hash so tampering is detectable.

Smaller items from the same review, as a checklist:
- [ ] `review-tracing.md` copilot example shows `gpt-5.6-sol` where the profile pins `gpt-5.4`
- [ ] `save_trace.sh` hardcodes `"executor": "claude-code"` in `run.meta.json` even for copilot runs
- [ ] Legacy `REVIEW_STATE.json` has no `reviewer_backend`; resume should default missing → `codex`
- [ ] `difficulty: hard` rebuttal-ruling phase has no copilot branch (only codex/manual)
- [ ] Routing doc top table lists `--reviewer: copilot` as an override for *all* reviewer skills, while the copilot section says only `/auto-review-loop` is wired — the two specs contradict
- [ ] Copy-mode install (and `smart_update_copilot.sh`) never deploys `.github/agents/` — copilot mode would then always be `REVIEW_UNAVAILABLE`
- [ ] No tests: profile deployment/reconcile/uninstall, routing fail-closed paths, legacy-state resume, trace backward-compat all uncovered (`tests/test_copilot_install.py` is a good place to extend)
- [ ] Codex-mirror routing doc: relative link resolves into the mirror itself; anchor target doesn't exist

Happy to re-review as soon as these are addressed — the doctrine-side asks are all mechanical once the Layer-1 contract is grounded in a live CLI session. Thanks again for the careful work here; this is one of the more thoughtfully structured community PRs we've received.

Functionhx:
Thanks for the detailed review. The Layer 2 items are clear.Will update the PR once everything is tested and aligned.

wanshuiyin (changes requested):
Thanks for the extremely thorough turnaround — I re-checked all of round 1's findings against the current diff, and every single one is genuinely fixed, several of them exceeding what I asked for:

- `COPILOT_CLI` auto-detection dropped entirely; explicit `--reviewer: copilot` opt-in only, with an honest citation of the actual upstream state (github/copilot-cli#2107 is still just a proposal).
- `.agent.md` extension used correctly and propagated into the copy-installer's manifest logic.
- The invocation mechanism now uses `copilot --agent NAME --prompt "..."` — I independently verified against the official Copilot CLI docs in round 1 that this *is* the documented programmatic form (`/agent`, natural-language, inference, or `copilot --agent NAME --prompt "..."`), so this is the right fix, not just a plausible-looking one.
- Copilot backend is explicitly "drive-only, not acquit" with `effort_unpinned: true` recorded in trace metadata — closes the review-floor bypass.
- `REVIEWER_MEMORY.md` is now append-only/verbatim/executor-forbidden-to-summarize, runs regardless of `REVIEWER_DIFFICULTY`, and you added a SHA-256 pre-call hash (`--memory-hash` on `save_trace.sh`) I hadn't even asked for — nice, matches the "receipts must be independently checkable, not just claimed" pattern this repo already uses elsewhere.
- Trace `executor` field reads from an `EXECUTOR` env var now instead of being hardcoded to `claude-code`.
- The routing-table self-contradiction, copy-install profile deployment, and the hard-mode rebuttal-ruling phase's missing copilot branch are all fixed.

I ran the full suite on your branch (493 passed, 16 skipped, 0 failed) and the inventory check — clean.

I had a second, fresh model re-review the whole thing (not just the diff) given the scope of changes, specifically told to try to poke holes in what I'd already confirmed rather than re-litigate it. It found three more real issues I verified myself by reading the actual current file state, not the diff hunks:

## 1. Shell injection via `REVIEWER_MEMORY.md` content in the round-2+ prompt template (`skills/auto-review-loop/SKILL.md` ~line 408, also ~line 233)

The hard-mode rebuttal-ruling template shows:
```bash
copilot --agent "<saved-reviewer-profile>" --prompt "[Rebutal ruling — same reviewer]

## Your Memory From Previous Rounds
[Paste full contents of REVIEWER_MEMORY.md]
...
```
This instructs the executor to construct the actual shell command by inlining `REVIEWER_MEMORY.md`'s raw content — which is now (correctly, per the fix above) the reviewer's own free-text response, verbatim — directly inside a double-quoted `--prompt` argument. A `"` in that text closes the argument early; a `` ` `` or `$(...)` gets re-interpreted by the shell. Since this is a template an LLM executor follows literally to build a real command it then runs, this is a genuine injection surface, not just a documentation nit. The safe pattern this repo already uses elsewhere (e.g. `save_trace.sh`'s python heredocs) is to pass untrusted content via env var or a temp file, never inline string interpolation into a quoted shell argument — e.g. write the assembled prompt to a temp file and use `--prompt "$(cat "$PROMPTFILE")"` (command substitution captures the file's bytes as one argument value; it doesn't re-parse them as shell syntax) or check whether `copilot` supports reading the prompt from stdin/a flag instead of an inline arg.

## 2. `reviewer_backend` conflation between "what just ran" and "what should run next" (`skills/auto-review-loop/SKILL.md` ~line 373 / Phase E)

When a copilot positive verdict triggers the mandatory cross-family escalation, the gate does `Update reviewer_backend in REVIEW_STATE.json to codex ... for the escalation round`. If Phase E's documentation step reads `reviewer_backend` to label the *current* round's result, this looks like it could mislabel the copilot-driven round's data as `backend=codex` before codex has actually run. Worth double-checking Phase E reads whichever backend value was active *during* the round being documented, not the value just written for the *next* round.

## 3. Cross-family escalation doesn't check the executor's own family (same location)

The escalation logic unconditionally schedules "codex (or manual if codex unavailable)" without checking the CURRENT executor's family first. This is fine when the executor is Claude (mainline pack) — codex is always cross-family there. But if this logic (or something structurally similar) ever runs under an OpenAI-family executor, "escalate to codex" would be a same-family reviewer, silently defeating the "mandatory cross-family acquittal" guarantee the text promises. Worth an explicit executor-family check before picking the escalation backend, or a comment making clear this logic assumes a Claude-family executor and is intentionally scoped to the mainline pack only.

Also noted, lower priority: `skills-codex/auto-review-loop/SKILL.md`'s Phase B.5.1 (touched by this PR only for the unrelated `run_id`/append-only-log fix, so likely pre-existing, not introduced by you) writes an acquittal line directly in the stop-condition step AND Phase E's own "if score >= 6" check can also fire — unlike the mainline's Phase B.5.1, which explicitly says "do NOT write here, Phase E documents" specifically to avoid this. Possibly worth the same one-line guard in the mirror while you're in there, though I don't think it's this PR's responsibility to fix pre-existing mirror behavior.

None of this takes away from how solid the rest of this turned out — the Layer-1 technical contract is now genuinely correct (and tested against real docs), and the doctrine fixes are careful and in most cases exceed what I asked for. It's really just the injection issue (#1) that I'd call a hard blocker; #2/#3 are worth a careful look but I'm not 100% certain from reading alone that they manifest exactly as described — a quick trace through the actual state transitions would settle it either way.

Functionhx:
Thanks for the very thorough review and for running the full test suite. The remaining issues are clear, especially the shell-injection risk. I’ll address them and update the PR soon.

wanshuiyin:
Round-2 verified — this is a thorough response, and it holds up under a fresh cross-model pass. All three round-1 blockers are confirmed closed:

1. **Shell injection** — dynamic memory/rebuttal content now reaches the prompt via `cat -- "$FILE"` into a temp file, never inside a heredoc body; the heredocs carry only static text, so even a memory line matching an `ARIS_*` terminator is just data.
2. **Backend field conflation** — resume reads the forward `reviewer_backend`, the current round snapshots to `round_backend`, the stop gate reads only the snapshot, and legacy state files get explicit `codex`/`false` defaults.
3. **Executor-family routing** — opposite-family profile forced from the derived executor family, `unknown` fails closed to `REVIEW_UNAVAILABLE`, and the profile's frontmatter model is re-parsed and family-checked before use.

The two new helpers are the right shape and, importantly, honest: `copilot_native_evidence.py` re-derives both families from host-persisted events, binds lifecycle/toolCallId/nonce/response, and rejects same/unknown-family pairs — while explicitly documenting that a same-UID local process could fabricate the events file and that this is "not cryptographic signing against a malicious local user." That residual trust is labeled, not hidden, which is exactly the standard this repo asks for (`identity_assurance: caller_declared` / `independence_verified: unverified` where applicable). `review_gate.py` upholds the one invariant that matters: no positive verdict stops the loop without re-validated cross-family evidence or the documented external finalizer; compatibility-mode Copilot positives can only escalate.

Pushed one cosmetic fixup directly (`69b51d5`): two prompt templates still said "The author (Claude)" — with the native route the executor is whatever the host runs, and the routing/stop gate already treat it that way.

584 tests green (+60 over main). Thanks for the persistence across two substantial rounds — this closes #258 properly. Merging.

PR #361, opened 2026-07-13
docs: sync research-lit source options

3mom3:
## Summary

- Document Gemini and OpenAlex as explicit research-lit sources.
- Sync the English and Chinese README, Zotero integration docs, and Codex mirror.

## Validation

- `python tools/check_skills_inventory.py`
- `python -m pytest tests/test_openalex_fetch.py tests/test_codex_skill_mirror.py`
- Result: 24 passed

Refs #240.


wanshuiyin (changes requested):
Thanks — the substance here checks out against the mainline source of truth (`gemini`/`openalex` are indeed explicit opt-in sources excluded from `all`, and the Zotero-doc rewrite into "default web tier vs opt-in external sources" is a genuine clarity win). Two things need fixing before merge; requesting changes so they don't get lost:

1. **The Codex mirror gains the table rows but not the execution steps.** `skills/skills-codex/research-lit/SKILL.md` now lists Gemini (row 8) and OpenAlex (row 9) as selectable sources, but the mirror's Step 1 has no corresponding invocation sections — it goes from the Exa block straight to PDF download. The mirror is a full workflow doc (it carries complete `$EXA_FETCHER` / `$DEEPXIV_FETCHER` bash blocks), so a Codex user passing `— sources: openalex` would have the source *accepted* but possibly never *invoked*. Please port the two execution sections from the mainline `skills/research-lit/SKILL.md` (adapting the resolver chain to the Codex variant, same as the existing S2/DeepXiv/Exa blocks do).

2. **Stale counts in the collapsed-section summaries.** After adding the two override examples, `README.md`'s `<summary>` still says "12 override examples" and `README_CN.md`'s still says "8 个 override 示例" — please bump both to the actual post-change counts (and double-check the "16/15 inline parameters" numbers while you're there).

Everything else is accurate and EN/CN-equivalent. Happy to merge once these two land — and #362/#363 look good as-is.

wanshuiyin (approved):
Both requested fixes confirmed: the Codex mirror now has full execution blocks for both new sources (Gemini MCP priority-1 / CLI-fallback priority-2, matching mainline's own pattern; OpenAlex fetcher resolution + call), and the override-example counts in both READMEs are corrected (EN 12→14, CN 8→10). Re-ran the full suite (486 passed, 16 skipped) and the inventory check on your branch — clean. Thanks for the quick turnaround.

PR #362, opened 2026-07-13
test: cover OpenAlex fetch helper

3mom3:
## Summary

- Add focused mock coverage for the OpenAlex fetch helper.
- Cover abstract reconstruction, metadata normalization, request filters, page-size limits, and rate-limit errors.

## Validation

- `python -m pytest tests/test_openalex_fetch.py`
- Result: 4 passed

### DIFF
--- a/tests/test_openalex_fetch.py
+++ b/tests/test_openalex_fetch.py
@@ -103,13 +103,17 @@ def test_search_works_builds_filters_and_caps_page_size(monkeypatch):
 
     assert captured["url"] == "https://api.openalex.org/works"
     assert captured["timeout"] == 30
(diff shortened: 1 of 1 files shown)


wanshuiyin:
CI failed on both platforms because the workflow installs only pytest while `tools/openalex_fetch.py` imports `requests` at module load — your tests are correct, the env just lacks the optional dep. Pushed a one-line `pytest.importorskip("requests")` to your branch (maintainer edit) so the module skips there instead of failing; locally with requests installed all 4 still run and pass. Will merge once CI re-runs green — thanks for the solid mock coverage, especially the filter-order decoupling in your second commit.

PR #371, opened 2026-07-15
Add citation-crosscheck skill: double-blind bib-vs-web field audit

dty0606:
A new skill that verifies bibliographic FIELD accuracy (not just existence) adversarially: agent A extracts each cited entry verbatim from the local .bib, agent B independently re-fetches the same paper from the web while blind to the .bib, and agent C diffs them field-by-field (author set+order, year, venue, arXiv-id/DOI). Emits a side-by-side .xlsx with MATCH/MINOR/MISMATCH per entry.

Complements /citation-audit (context-fit); this owns field accuracy. A NEEDS_WEBCHECK route avoids false-condemning legitimate non-arXiv (blog/ tech-report) citations the web agent's index search can't see; the main agent web-confirms every proposed fix before any .bib edit.

README: adds a What's New entry.


wanshuiyin:
Thanks for this — the double-blind bib-vs-web design (two agents that never see each other's output, verified against a neutral verifier, `NEEDS_WEBCHECK` routing so a non-arXiv citation isn't false-condemned) is a genuinely useful complement to `/citation-audit`'s context-fit focus. Ran the repo's mechanical gates locally against this branch and it currently fails four of them — flagging so you don't have to rediscover each one from a red CI run:

**1. Missing Codex CLI mirror** (`python3 tools/check_skills_inventory.py`)
Every mainline skill under `skills/` needs a counterpart under `skills/skills-codex/` — this repo ships two independent SKILL.md sets (Claude Code + Codex CLI conventions) kept in lockstep. `skills/skills-codex/citation-crosscheck/SKILL.md` doesn't exist yet. The main behavioral thing to translate: mainline Step 2 uses Claude Code's `Agent` tool for the A/B parallel launch — the Codex mirror convention for existing fan-out skills (e.g. `skills/skills-codex/research-lit/SKILL.md`) is a good reference for how that gets expressed for Codex CLI.

**2. Missing catalog entry + stale skill counts** (same check)
`docs/SKILLS_CATALOG.md` needs a `citation-crosscheck` entry, and the skill-count literals in `README.md`, `README_CN.md`, `AGENT_GUIDE.md`, `docs/ARIS_INTRO.md`, `docs/ARIS_INTRO.html`, `skills/skills-codex/README.md`, `skills/skills-codex/README_CN.md` all need to go from 80 → 81.

**3. Missing `tools/skill-groups.tsv` entry** (`pytest tests/test_skill_groups.py`)
New skills need a group assignment in the selective-install catalog — `test_every_upstream_skill_is_cataloged` fails without it.

**4. `Agent` grant without a fan-out-pattern citation** (`check_skills_inventory.py`'s advisory lint)
The SKILL.md frontmatter grants `Agent` (used for the Step 2 A/B parallel launch + Step 3 verifier), but the body doesn't cite `shared-references/fan-out-pattern.md`. Every other skill in this repo that fans out cites it — worth a one-line pointer in your Step 2/3 description. (Heads up: PR #372, open right now, is formalizing a stricter "bounded fan-out contract" — fixed shard/concurrency/turn caps + a mandatory read-only leaf agent — for the three skills that currently fan out. Once that lands, citation-crosscheck's A/B/C launch would likely need to adopt the same contract to stay consistent; worth a look once #372 merges, though not a blocker for this PR today.)

Once `tools/check_skills_inventory.py` and `pytest tests/test_skill_groups.py tests/test_codex_skill_mirror.py` are clean locally, this should be in good shape — happy to take another look.

wanshuiyin:
Thanks for the fast scaffolding turnaround — mirror, catalog entry, count bumps, tsv entry, and the fan-out cite all landed, and the mechanical gates (inventory + `test_skill_groups` + `test_codex_skill_mirror`) now pass locally against this branch. Your main-merge also came through clean (no mainline regressions).

Ran the substantive re-check the gates can't cover (fresh cross-model pass). Two issues need another round — both about the skill's core invariant, so flagging them as blocking:

**1. The Codex mirror claims a blindness guarantee it can't deliver as written.** The mirror says B runs via fresh `spawn_agent` (or "sequential execution in fresh contexts" without delegation) and asserts blindness is *identical* across hosts. But the main agent has already read the `.bib` in Step 1, and nothing in the written flow forbids the spawned/sequential context from inheriting that parent context — stock `spawn_agent` context-forking would hand B the very `.bib` it must never see, and "fresh sequential contexts" has no enforcement mechanism or fail-closed path behind it (the frontmatter doesn't even declare `spawn_agent`). This isn't an honest degradation — the mirror wraps a possibly single-context check in a double-blind certification. Two acceptable fixes: (a) specify an isolation mechanism that demonstrably does not inherit parent history, passing B only the title+first-author list, with `BLOCKED` when isolation can't be provided; or (b) honestly downgrade the Codex flow to a labeled non-blind heuristic that can flag but never certify. Either is fine; claiming blind-equivalence without enforcement is not.

**2. Internal contradiction in the blindness rules — in both mainline and mirror.** Step 2 restricts B to the title+first-author list and forbids reading the `.bib`; the Key Rules then say "B must try the general web and **the entry's own `url`**." The `url` field lives inside the `.bib` entry — B structurally cannot know it without breaking the blindness the same section declares to be the mechanism. The fix is what Step 4 already implies: B tries only landing pages its own independent search discovers; the entry's own `url` stays exclusively with the main agent's Step-4 webcheck. One-line rewording in each file.

Non-blocking, worth fixing in the same push:
- `docs/ARIS_INTRO.md` (and the generated `.html`) still say the Audits category has 5 skills and don't list citation-crosscheck — should be 6 + regenerated HTML. The global count is correctly 82.
- The catalog lists only `web access` as a requirement, but Step 5 hard-depends on `openpyxl` for the `.xlsx` — declare it (or add a no-dependency CSV fallback).
- The fan-out-pattern cite is currently a bare reference: the A/B JSON has no `shard_id`/`dedup_key`, subagents write shared `/tmp` files directly, and there's no coverage verification that both sides processed all N citations. Tolerable for now given the fixed A/B/C structure caps concurrency, but note that PR #372 (bounded fan-out contract) is likely to land soon and this skill's launch pattern will need to adopt it — a follow-up is fine.

Also verified: no routing collision with the newly-merged `web-debug-search` (different trigger domains), and no maintainer personal info in the diff. The core design remains something we want — it's specifically the mirror's honesty about what it can guarantee that needs one more pass.

dty0606:
Beyond addressing the above comments, I squashed the three branch commits into one on top of current main (09f1f98) so the review round lands as a single clean commit; note this retired 78f1cda, so any inline comments there now point at the old SHA?

wanshuiyin (changes requested):
Round-2 review done. First: this rework is genuinely impressive — the conditional-blindness model (isolation attested or honest flag-only degradation), the fail-closed deterministic gate that recomputes the A/B diff itself instead of trusting shard C, and the 86-case selftest with CJK/accent/truncation coverage are exactly the right architecture. Both round-1 blockers are properly closed. The fresh cross-model pass (which actually *executed* the verifier's functions, not just read them) plus my own reproduction found a set of new issues in the rewrite — the theme is that a few paths still let shard output or lossy normalization reach MATCH without the gate's protection.

**Blocking:**

**1. Step-4 ladder has ESCALATE-bypass paths.** The ladder promises every ESCALATE gets a main-agent primary-source fetch, but the single-side-missing-field branch allows assigning MATCH/MINOR "from the fields both sides share" — shard-only evidence becoming a verdict with neither the gate nor a fetch (SKILL.md ~403, ~439). And the ladder is first-match-wins but internally contradictory: the URL-flag branch can conclude MISMATCH on a dead link alone (~416) while the later found:false branch requires the main agent's own search to also fail (~424, ~432) — the earlier branch preempts the stricter one. The ladder needs re-ordering so no branch terminates in MATCH/MINOR/MISMATCH without either (a) the gate's mechanical decision or (b) a completed main-agent primary-source fetch.

**2. Verified false-accepts in `cx_verify.py`'s normalization.** I reproduced these directly against the module:
- `author={{Research and Development Team}}` (one corporate author) and the unbraced two-person split normalize to the *identical* author list — brace grouping is stripped before `and`-splitting (~75, ~170), so a corporate-vs-people discrepancy can MATCH.
- `&` (and `;`) are treated as author separators (~190); BibTeX only recognizes `and` — `Alice Smith & Bob Jones` should be one malformed author, not two clean ones.
- Duplicate `author` fields in one entry: `setdefault` silently keeps the first and ignores the conflicting second (~517, ~527) — conflicting evidence should ESCALATE, not vanish.
- Title punctuation folding equates `Model 4.1` with `Model 4-1` (~83) — confirmed by direct call; for version-number titles that's a substantive difference. Consider keeping digit-adjacent punctuation significant.
False-escalates are safe; these are all in the dangerous direction. Each deserves a selftest case once fixed.

**3. C is still in the trust path for routing.** The gate takes `flags` exclusively from C — a C that omits `degraded_search` waves the entry through, and if C emits `flags` as a string instead of a list the gate silently coerces to `[]` instead of failing closed (~599, ~1073). Recompute whatever routing conditions are derivable from A/B raw artifacts (e.g. `web_record_absent` from B's own `found`, `url_only_source_unqueried` from B's `sources_tried`), and hard-fail on C schema violations.

**4. Coverage root has gaps.** With a `.bbl` present only counts are compared, not exact key sets (drop one real citation + gain one stray `.tex` cite = same count). `--expect`/`--expect-keys` are CLI-optional and validate ordinal *sets*, not the ordinal→cite-key/title mapping, so drop-and-substitute survives (~996, ~1043, ~978). Also `$CITED_KEY_COUNT` is used later in the SKILL (~377) but the earlier snippet only *prints* `CITED_KEY_COUNT=...` — it never sets a shell variable (~166). Suggest: a deterministic manifest (ordinal, cite key, title hash) that the gate verifies as a mapping, with `--expect-keys` made mandatory in the documented invocation.

**5. Codex mirror still overclaims unconditionally in its headline surfaces.** The conditional-isolation section you added is right, but the mirror's frontmatter description, title, method table, and `BLIND = true` constant (~3, ~49, ~69, ~76) still assert double-blind unconditionally — the description is what routes skill selection, so it must carry the conditional ("conditionally double-blind; flag-only heuristic when isolation can't be attested").

**6. Wire the selftest into CI — pre-merge, not follow-up.** Nothing in `tests/` or the workflow runs `cx_verify.py --selftest`; for a load-bearing gate that's a required check. One small pytest that (a) subprocess-runs `--selftest` expecting exit 0 and (b) asserts the two mirror copies are byte-identical would do it.

**Non-blocking:** shard text is printed unescaped (ANSI/newline log-injection surface, ~958, ~1081 — sanitize before printing); the helper's resolver uses a custom `skills/.../scripts` chain without the canonical `tools/` shim — either add the shim or note the owner-local exception explicitly per `integration-contract.md` (the Policy-A registration line itself is correct).

The architecture is right and none of this changes it — it's hardening the exact properties the rewrite was built to guarantee. Happy to do a fast round 3 once these land.

dty0606:
1. **Step-4 ladder** — reordered so every MATCH/MINOR/MISMATCH comes from either the gate or a completed main-agent fetch. A fabrication MISMATCH now needs the `url` dead *and* my own search to also
  fail, so no branch preempts the stricter one. Also fixed the stale "Key rules" line that still implied dead-`url`-alone.
  2. **`cx_verify` normalization** — brace-aware author split (a braced `{… and …}` stays one author), `;`/`&` no longer split, duplicate conflicting fields ESCALATE, and `4.1` ≠ `4-1` in titles. Each has
  a selftest case.
  3. **C out of the routing trust path** — `web_record_absent` / `degraded_search` / `url_only_source_unqueried` are recomputed from B's `found`/`sources_tried` + A's fields; C's flags are advisory only, 
  and a bad C schema fails closed.
  4. **Coverage** — a manifest (ordinal → cite-key → title-hash) is checked as a mapping (catches drop-and-substitute), coverage is now mandatory, `$CITED_KEY_COUNT` is actually set as a shell var, and
  the `.bbl` check compares key sets.
  5. **Codex mirror** — description, title, method table, and the `BLIND` constant now say *conditional* double-blind.
  6. **CI** — added `tests/test_citation_crosscheck_gate.py` (runs `--selftest` on both copies + asserts byte-identity); the existing suite picks it up.

dty0606:
Round-3 pushed — rebased onto current `main` (`94d8093`), squashed to one commit; merges cleanly now.

All six blockers done: Step-4 ladder (every verdict rests on the gate or a completed fetch), normalization false-accepts (brace-aware author split, `;`/`&` not separators, duplicate-field ESCALATE, `4.1`≠`4-1`), routing recomputed from A/B with C advisory-only + fail-closed, mandatory coverage manifest (ordinal→cite-key→title-hash), conditional Codex-mirror blindness, and `--selftest` wired into CI. Non-blocking items too. A self-run adversarial pass caught one extra false-accept (a suffix-token surname like `Ii`/`III` dropping a coauthor) — fixed. `--selftest` 113/113, full suite green locally on 3.12.

Heads-up: the workflows are in `action_required` — they need your approval to run on the fork PR. Ready for round 3 whenever.

PR #372, opened 2026-07-17
fix: bound recursive subagent fan-out with a read-only leaf agent

Dennis-Huangm:
## Motivation

`idea-creator`, `research-lit`, and `proof-checker` benefit from parallel work, but previously described fan-out independently without a shared execution boundary or a reliable record of which units were actually processed. In long research workflows, unconstrained nested delegation can multiply tasks, consume the watchdog budget, and leave partial coverage that the parent cannot distinguish from a complete run. This PR introduces a reusable bounded fan-out contract and a read-only Claude leaf agent while keeping research judgments, final verdicts, and shared-file writes in the parent.

### Read-only Claude leaf agent

Add `agents/aris-fanout-leaf.md` with a hard `Read`, `Grep`, and `Glob` tool allowlist. The leaf processes one assigned shard, cannot delegate or mutate files, and cannot render novelty, quality, relevance, or proof-validity verdicts.

### Canonical bounded fan-out contract

The mainline and Codex references define the same executor-managed limits: at most 8 shards, at most 4 active workers, an 8-turn worker budget where supported, no replacement worker after failure, one executor-side sequential fallback, and a mandatory mechanical coverage receipt.

The receipt records expected, completed, failed, covered, and uncovered shard and unit IDs. It proves execution coverage only; it does not prove candidate quality, paper relevance, novelty, proof validity, obligation-discovery completeness, or publication readiness.

### Honest Codex mirror semantics

The Codex mirrors now state explicitly that stock `spawn_agent` does not expose a per-child tool allowlist, recursion flag, or portable child turn cap. Read-only, non-recursive, and turn-bounded behavior is therefore a prompt convention unless the active host supplies structural enforcement; stock Codex does not provide the same hard worker boundary as the Claude leaf.

Shard count, concurrency, no-replacement behavior, and coverage accounting remain executor-managed and mechanically enforceable.

### Locally readable research evidence

`research-lit` now requires the executor to freeze one local read-only evidence bundle per shard before dispatch. Each paper record includes identity and verification fields plus available abstracts, excerpts, local PDF paths, local note paths, and an explicit evidence status.

Papers without locally readable evidence remain `UNPROCESSABLE_NO_LOCAL_EVIDENCE` and uncovered rather than asking the read-only leaf to infer methods or results from titles or remote URLs.

### Source-unit proof coverage

`proof-checker` now freezes mechanically enumerable source units—stable file spans, sections, theorem blocks, or proof environments—before dispatch. Proof obligations are discovered during extraction and receive deterministic IDs afterward.

The coverage receipt therefore proves source-unit execution coverage only and does not claim that every possible proof obligation was discovered or that the extracted ledger is mathematically complete.

### Owned Claude leaf lifecycle

The Claude installers record leaf ownership in a separate project-bound `.aris/installed-agents.txt` manifest only when ARIS creates the link.

The Bash and PowerShell installers now:

- create and record the leaf before committing a fan-out skill installation;
- reuse an exact installer-owned link;
- use but never adopt or delete a pre-existing matching external link;
- retain the leaf while any fan-out skill remains installed;
- remove an owned leaf after the final fan-out skill is reconciled away;
- preserve retargeted, replaced, or externally linked paths;
- support agent-only cleanup retries;
- narrowly roll back a leaf and ownership record created by a failed invocation;
- validate the ownership record against the current project before mutation.

Codex-only installations do not create or inspect the Claude agent manifest.

### Legacy `aris` skill compatibility

A verified legacy `.claude/skills/aris` link pointing to the old nested skills directory can now be migrated when a future or custom repository contains a genuine skill named `aris`. The path is revalidated immediately before removal, and dry-run remains non-mutating.

### Windows CI coverage

Changes under `agents/**` now trigger the Windows installer workflow for both Windows PowerShell and PowerShell Core.

## Safety properties

- At most 8 planned shards per fan-out stage.
- At most 4 executor-dispatched workers per wave.
- At most one sequential fallback per failed shard.
- No replacement worker after failure.
- Claude leaf workers cannot delegate, run shell commands, or write files.
- Missing units remain explicit in the coverage receipt.
- Ownership is never inferred from target equality alone.
- User-owned and retargeted leaf paths are preserved.
- Removal is limited to the exact link recorded as installer-owned.
- Failed skill application does not leave a newly created untracked leaf.

## Testing

Local verification for the review-fix commits:

- bounded contract, Codex mirror, and related contract tests: 28 passed;
- ARIS skill inventory: consistent;
- Bash selective installer lifecycle under WSL: 28 tests passed, 1 environment-dependent skip;
- Bash tools and leaf lifecycle under WSL: 13 passed;
- Windows PowerShell 5.1 installer suite: 22 passed, 7 file-symlink tests skipped because the local machine lacks symbolic-link privilege;
- `bash -n tools/install_aris.sh`: passed;
- `git diff --check`: passed.

The full GitHub Actions matrix will run after these commits are pushed.

## Non-goals

- No service-preflight redesign.
- No review-trace schema redesign.
- No research Manifest stage redesign.
- No proposal-ready/evidence-ready state redesign.
- No general installer transaction redesign.
- No automatic migration between different ARIS repository locations.
- No deterministic dispatcher that forces the parent model to invoke a leaf.


wanshuiyin (changes requested):
Thanks for this — the shared bounded-execution contract (fixed shard/concurrency/turn caps, a real tool-restricted leaf agent on the Claude side, mandatory coverage receipts honestly scoped to execution-only) is exactly the right shape for closing the "nested fan-out with no record of what was actually covered" gap. Ran the mechanical gates locally against this branch first: `python3 tools/check_skills_inventory.py` → consistent, full `pytest -q` → 533 passed / 26 skipped / 0 failed. All green.

Sent it through an independent cross-model review pass on top of that, since this changes core doctrine (`fan-out-pattern.md`) and installer file-system mutation logic that every ARIS install goes through. That pass came back **request-changes** on substance the local checks can't catch. I verified the two I could quickly confirm myself by reading the code directly; the rest are relayed with their file:line citations for you to check independently.

## Blocking

**1. Codex mirror doesn't actually enforce what it describes** (`skills/skills-codex/{idea-creator,research-lit,proof-checker}/SKILL.md`)
`shared-references/fan-out-pattern.md`'s Codex section is appropriately hedged — it says the mirror "does not pretend to install or invoke" the Claude custom agent, and uses a "leaf-equivalent prompt" instead. But the three individual skill mirrors then say things like "every worker prompt is leaf-equivalent, read-only, and declares `recursion: false`" (`idea-creator/SKILL.md` fan-out contract section) without repeating that hedge — read in isolation, that phrasing suggests `recursion: false` / `max_turns: 8` are real `spawn_agent` invocation parameters. Standard Codex CLI's `spawn_agent` has no such parameters, no per-child tool allowlist, and no recursion cap — a spawned Codex shard can still delegate further. On the Claude side the leaf's `Read/Grep/Glob`-only tool allowlist + `maxTurns: 8` (`agents/aris-fanout-leaf.md`) is a real structural boundary; on the Codex side the same limits are prompt convention only, and the individual mirror files don't make that distinction visible to a reader who doesn't also read the shared reference. Worth either re-stating the hedge in each mirror skill, or having those skills explicitly point back to the shared-reference's honest framing.

**2. Leaf-agent lifecycle has no ownership tracking — confirmed live bug, not a mirror-honesty nit**
I checked this one directly: `REQUIRES_LEAF_AGENT` (`tools/install_aris.sh:1251`) only ever gates a *positive* `ensure_leaf_agent` call (lines 1263, 1285) — there's no corresponding branch that removes the leaf when reconcile drops the last of idea-creator/research-lit/proof-checker from the selected set. The only removal path, `remove_leaf_agent_on_uninstall` (line 1037), is wired solely into `do_uninstall` (full uninstall, line 1159) — not into the reconcile path. So: reconcile away the last fan-out skill → `.claude/agents/aris-fanout-leaf.md` symlink lingers with nothing in the manifest recording that this installer put it there. It also means uninstall can't distinguish "a leaf this installer created" from "a user's own symlink that happens to point at the same source file" — there's no manifest/sidecar entry for it, unlike the existing `.aris/tools` symlink pattern this PR is otherwise modeled on. Needs a manifest entry for leaf ownership, plus a reconcile-side removal branch mirroring the existing add branch.

**3. `research-lit`'s leaf can't necessarily see what it's asked to analyze**
The leaf agent has *only* `Read/Grep/Glob` — no WebFetch/WebSearch/Bash/MCP (`agents/aris-fanout-leaf.md`). But `research-lit` doesn't download PDFs by default, and the candidate artifact it writes to disk (SKILL.md's Step-1 output, `verify_papers.py`'s output) carries only identity fields — ID, DOI/arXiv-id, title — no abstract or source excerpt. The leaf is then asked to extract method/results/relevance per shard, information that typically only exists in the parent's search context or a remote page the leaf has no tool to fetch. Worth checking whether dispatch needs to freeze a local evidence bundle (abstract/excerpt/local-PDF-path per paper) before handing shards to the leaf — otherwise the default (no-download) research-lit path could hand the leaf shards it structurally cannot complete.

## Worth fixing, not blocking on its own

**4. Install ordering isn't atomic w.r.t. the leaf requirement.** Bash applies the skill plan → commits manifest → *then* creates the leaf (`install_aris.sh:1277`, `1283`); a failure creating `.claude/agents/` or the symlink after that point leaves a project with the fan-out skill installed but its required worker missing. PowerShell does the reverse order (leaf first, then skill apply + manifest commit) — a failure there leaves an orphaned, unrecorded leaf. Same root cause as #2 (no ownership record) — worth fixing together.

**5. `proof-checker`'s coverage-unit definition has a timing circularity.** It calls for freezing the complete section/theorem/**obligation** ID set before extraction (`proof-checker/SKILL.md`), but Phase 0.5 *is* the extraction phase that discovers obligations — an obligation can't get a canonical ID before the phase that finds it runs. So `coverage_complete: true` can honestly prove every pre-declared section/theorem was processed, but not that every obligation was extracted — yet the text calls the result a "complete merged ledger." Suggest freezing only the mechanically-enumerable source-span/section/theorem IDs pre-extraction, generating obligation IDs post-extraction, and scoping the receipt's claim to source-unit coverage rather than ledger completeness.

**6. Stale doctrine text.** `fan-out-pattern.md` (and its mirror) still carries a warning that idea-creator does "same-family quality filtering" in Phase 3 — Phase 3 is now mechanical dedup + an objective budget gate (`idea-creator/SKILL.md`). Doc-only, but worth cleaning up in the same pass.

**7. `migrate_legacy` reordering** — moved from immediately-after-legacy-detection to right before the dry-run-exit / apply points. Confirmed no regression against the *current* repo state (there's no upstream skill literally named `aris`, so `compute_plan` never observes the legacy path either way). But it's not general behavior-preservation: if a future or custom ARIS repo ever ships a skill named `aris`, the new ordering would let conflict-resolution see the not-yet-migrated legacy symlink and misclassify it as a foreign conflict against that skill. Low priority given today's tree, but worth a comment noting the assumption, or migrating that one path before plan computation regardless of dry-run/apply branch.

Nothing here touches the core idea — the contract shape and the honest execution-only scoping of the coverage receipt are right. It's the two lifecycle gaps (#2/#4) and the mirror-honesty/reachability gaps (#1/#3) that I'd want closed before merge, since #2 is a concrete bug I confirmed by reading the installer directly, not a hypothetical.

wanshuiyin:
Re-reviewed the three fix commits (`5f2d6e0`, `0b217f1`, `c988b08`) with a fresh cross-model pass, verifying each of the 7 findings from the last round against the current tree. Result: **all 7 are closed** — and several went beyond what was asked:

1. **Mirror overclaim — closed.** All three Codex mirrors now state explicitly that `max_turns`/`recursion: false`/tool limits are prompt convention on stock `spawn_agent`, and the shared reference scopes the turn cap to hosts that expose one.
2. **Leaf ownership/lifecycle — closed, thoroughly.** The new `.aris/installed-agents.txt` ownership manifest with the `CREATE_OWNED / REUSE_OWNED / REUSE_EXTERNAL / REMOVE_OWNED / RETIRE_OWNERSHIP / RELEASE_OWNERSHIP` state machine is exactly the right shape — ownership never inferred from target equality, external leaves preserved, reconcile removal wired for the required=false path.
3. **research-lit leaf reachability — closed.** Mandatory per-shard local evidence files with an `UNPROCESSABLE_NO_LOCAL_EVIDENCE` honest-failure path, mirrored on the Codex side.
4. **Install ordering — closed.** Both installers now run prepare-leaf → apply-skills → commit-manifest → finalize-removal with rollback on pre-commit failure, symmetric across bash/PowerShell.
5. **proof-checker circularity — closed.** Source-unit IDs frozen pre-extraction, obligation IDs derived post-extraction, receipt honestly scoped.
6. **Stale doctrine — closed** on both sides.
7. **migrate_legacy — closed** beyond the ask: a real pre-plan special-case instead of just a comment.

One **new blocking issue** the fixes introduced, which I confirmed directly in the code:

**`leaf_target_resolves_to` doesn't canonicalize absolute symlink targets** (`tools/install_aris.sh` — the function reads the link target and, for absolute targets, compares raw strings against `$LEAF_AGENT_SOURCE`). `c988b08` canonicalized the ownership `project_root` but not this comparison. On macOS this bites reproducibly: `/tmp` is itself a symlink to `/private/tmp`, so installing with the repo at one alias and reconciling via the other yields physically-identical but string-different targets → spurious `CONFLICT` on a perfectly healthy owned leaf, breaking re-entrant reconcile. The new tests cover a *project*-path alias but not a *repo*-path alias, which is why this slipped through. PowerShell has the sibling issue: `Same-Path` does `GetFullPath` + case-insensitive compare but never uses the `Resolve-ReparseChain` helper that already exists in the script, so junction/reparse aliases on the repo path would misclassify the same way.

Suggested fix shape:
- Accept a raw exact match first (this preserves dangling-link cleanup when the repo has been deleted — don't lose that);
- If raw differs, canonicalize both sides and treat them as equal only when both canonicalizations succeed and match;
- Add the "install via real path, reconcile/uninstall via repo alias" test on both bash and PowerShell;
- Switch the PowerShell ownership `project_root` comparison to the reparse-chain-aware helper while you're there.

That's the only thing between this and merge — the hard parts (ownership model, atomicity, honest mirrors) are done, and done well. Also re-checked the fix commits for maintainer personal info: clean.

Dennis-Huangm:
Thanks for the follow-up, and apologies for the delayed response.

I addressed the remaining repository-alias blocker in `18aff77` and merged the current `main` in `22f10d8`. Path ownership checks now use raw-first, physical-second identity on both Bash and PowerShell, so symlink/Junction aliases are recognized without losing cleanup support for installer-owned dangling links.

I also added regressions covering repo-alias reconcile/uninstall, shared tools ownership, project aliases, dangling managed links, and independent leaf/profile ownership manifests.

The updated GitHub Actions matrix is green:
- Ubuntu pytest
- macOS pytest
- Windows PowerShell 5.1
- Windows `pwsh`
- skill inventory and advisory lint

Thanks again for the detailed review. The PR is ready for another look.

PR #376, opened 2026-07-23
Fix Codex experiment-queue adapter and helper resolution

yotta2020:
## What changed

- bundle the canonical `experiment-queue` helpers in the Codex skill mirror
- detect detached GNU Screen sessions reliably
- honor project-managed queue adapters and fail closed when required adapter files are incomplete
- prefer helpers shipped with project/global Codex skill installs, with canonical and legacy fallbacks
- add regression coverage for helper parity, install modes, resolver precedence/fallback, and screen detection

## Why

Codex project installs could resolve queue helpers through the mainline path instead of the bundled mirror, and a partially configured project adapter could silently fall back to the generic queue workflow. Both behaviors could bypass project-specific safety and recovery contracts.

## Validation

- `uv run --with pytest python -m pytest tests/test_codex_skill_mirror.py -q` — 25 passed
- Python syntax checks passed for both queue managers and the Codex manifest builder
- `git diff --check` passed
- adversarial review: P0/P1/P2 = 0

No paper template commit or unrelated working-tree deletion is included.

### DIFF
--- a/skills/experiment-queue/SKILL.md
+++ b/skills/experiment-queue/SKILL.md
@@ -119,6 +119,33 @@ A "wave" is a batch of jobs that fit available GPUs. Next wave only starts when:
 
 ## Workflow
 
+### Step 0: Honor a project-managed queue adapter
+
+Before parsing a manifest or running any generic pre-flight, resolve the project
+root and check whether it provides all of these files:
+
+- `CLAUDE.md`
+- `docs/ARIS_PROJECT_LAYOUT.md`
+- `code/scripts/ttad_aris_queue_wrapper.py`
+
+If they exist and `CLAUDE.md` declares an ARIS remote mapping with
+`queue_status`, this repository has a project-managed queue adapter. Read its
+root `AGENTS.md`, `CLAUDE.md`, `docs/ARIS_PROJECT_LAYOUT.md`, and
+`docs/GPU_REMOTE_WORKFLOW.md` completely before acting.
+
+- If `queue_status` is not `active`, stop and report the documented blocker.
+- If it is `active`, follow the project's wrapper workflow exactly. For the
+  TTAD adapter this means selecting/locking GPUs with `./check_gpu.sh
+  --gpu --select`, preparing with
+  `code/scripts/ttad_aris_queue_wrapper.py prepare`, and supervising the remote
+  canonical manager with the same wrapper's `supervise` command.
+- Do not continue into generic Steps 1-5. In particular, do not scan or select
+  GPUs independently, accept generic arbitrary `cmd` fields, launch a raw
+  `queue_manager.py`, or treat a pre-existing expected output as completion.
+
+The remaining workflow is only for projects that do not declare a
+project-managed queue adapter.
+
 ### Step 1: Parse Manifest / Build from Grid
 
 Input can be:
--- a/skills/experiment-queue/SKILL.md
+++ b/skills/experiment-queue/SKILL.md
@@ -122,16 +122,20 @@ A "wave" is a batch of jobs that fit available GPUs. Next wave only starts when:
 ### Step 0: Honor a project-managed queue adapter
 
 Before parsing a manifest or running any generic pre-flight, resolve the project
-root and check whether it provides all of these files:
+root and inspect `CLAUDE.md`. If it declares `queue_status`, the project has
+opted into a managed queue adapter. This declaration is authoritative: do not
+fall back to the generic queue when adapter files are incomplete.
 
+Require all of these files:
+
+- `AGENTS.md`
 - `CLAUDE.md`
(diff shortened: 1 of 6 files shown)


wanshuiyin (changes requested):
Thanks for this — the parity discipline here is genuinely good: I verified the new 428-line Codex `queue_manager.py` and `build_manifest.py` are byte-identical copies of the canonical files with strict parity tests guarding against drift, and the screen-detection regex change is correct (exact `PID.session-name` match, state-independent, multi-session safe — verified against Detached/Attached/multi/prefix/empty cases).

That said, this touches the resolution order for code that gets **copied to a GPU server and executed**, which we treat as security-sensitive by default. Ran the full cross-model pass and there are two blocking issues plus one design question:

**1. [Blocking] The new resolver preference lets any checked-out project silently hijack remote execution.** The Codex SKILL's resolver picks `.agents/skills/experiment-queue/scripts` *first*, based purely on two files existing — no realpath check against the install manifest, no hash, no user confirmation. The chosen `queue_manager.py` is then rsync'd to the GPU server and executed. Net effect: cloning any repository that ships a same-named script pair is sufficient to replace the canonical scheduler with repo-controlled code on your GPU machines. `.aris/installed-skills-codex.txt` can't serve as the trust root either — it's equally project-controlled. The new tests actually lock this precedence in (they simulate a trusted symlink and an incomplete-dir fallback, never a malicious complete dir). Fix shape: only prefer the project copy when its resolved realpath matches `repo_root` + expected relative path from the install manifest; otherwise fall through to global/canonical, or require explicit confirmation.

**2. [Blocking] "Fail closed" currently exists only as prose, and the Codex mirror gates on the wrong file.** The detect-adapter → validate-files → stop flow is natural-language instruction with no deterministic preflight (shell or Python) behind it, and the test named for failing closed only asserts string presence in the SKILL.md — it cannot demonstrate that a missing adapter file actually blocks the generic fallback. Separately, the Codex mirror still keys adapter detection off `CLAUDE.md`; a Codex-native project declaring its adapter in `AGENTS.md` sails straight into the generic workflow, which contradicts the mirror convention. Fix shape: a small deterministic preflight script + subprocess-level tests for {active+complete, active+missing-file ⇒ nonzero exit, inactive ⇒ block, AGENTS.md declaration, no-adapter ⇒ generic allowed}.

**3. [Design question] The adapter protocol is hardcoded to one private project's layout.** Any project declaring `queue_status` is required to ship `code/scripts/ttad_aris_queue_wrapper.py`, `docs/ARIS_PROJECT_LAYOUT.md`, `docs/GPU_REMOTE_WORKFLOW.md`, and `./check_gpu.sh` — the TTAD-specific wrapper and doc names look ported from your own project's internal contract. A non-TTAD project opting into a managed adapter would be wrongly blocked by these requirements. Could you confirm the TTAD layout is intentionally public, and generalize the protocol so the *project* declares its required files/entrypoint (with your TTAD layout as the worked example) rather than the skill hardcoding one layout for everyone?

The underlying problems this PR fixes are real (mainline-path leakage from Codex installs, silent generic fallback) and worth fixing — the parity machinery can stay exactly as is. It's the trust boundary of the resolver and the enforceability of the fail-closed claim that need one more round.

wanshuiyin:
Heads-up on rebasing: the canonical `queue_manager.py` has moved under you, in a way that should make this PR smaller rather than larger.

[#393](https://github.com/wanshuiyin/Auto-claude-code-research-in-sleep/pull/393) landed your `screen_exists()` fix verbatim — same `run("screen -ls")` + anchored-regex body — because it was correct and the scheduler was killing healthy jobs at the first 60s poll without it. Thank you for that one.

It also fixed four further defects in the same file, which matters for your mirror copy:

- The `pid dead + screen alive` branch had a comment and no `return`, so it fell through to `return "running"` forever. That branch was unreachable while the TAB bug existed — your fix makes it live, so fixing only `screen_exists` traded one hang for another.
- `failed_other` had no transition out and was not in `all_done()`'s terminal set; it is now parked as `stuck`.
- `output_exists()` ran `ls` with `shlex.quote()` on the whole path, which prevents glob expansion — the wildcard `expected_output` in our own manifest examples never matched.
- `depends_on` given as a bare string (the shape both SKILL.md examples showed) was iterated character by character, so a dependent phase never became ready.

So when you rebase, your `queue_manager.py` hunk becomes a no-op merge, and your byte-equality parity test should pass against an already-fixed canonical file instead of needing the mirror copy to reproduce those four defects.

The two blocking items from the 2026-07-30 review are unchanged and still yours: the resolver preferring a project-local script pair without checking it against the install manifest (that file is rsync'd to the GPU host and executed), and `fail-closed` existing only as prose with a test that asserts string presence. The parity machinery itself can stay exactly as it is.

PR #377, opened 2026-07-24
fix: tolerate unsupported picker cursor visibility

Dennis-Huangm:
## Summary

- keep the curses checkbox picker usable when a terminal supports drawing but rejects cursor visibility changes
- treat `curses.curs_set(0)` as an optional visual enhancement while preserving the existing fallback for other curses failures
- add a focused regression test that reproduces the `curs_set() returned ERR` behavior observed in Windows Git Bash

## Motivation

On some Windows Git Bash terminals, the interactive installer currently abandons the full checkbox picker and falls back to per-group prompts because hiding the cursor returns `ERR`. Cursor visibility does not affect picker functionality, so this cosmetic capability should not disable the otherwise working interface.

## Testing

- `PYTHONUTF8=1 PYTHONIOENCODING=utf-8 python -m pytest tests/test_skill_picker.py -q` — 9 passed, 3 skipped
- `git diff --check`
- manually verified the installer checkbox interface in Windows Git Bash

### DIFF
--- a/tests/test_skill_picker.py
+++ b/tests/test_skill_picker.py
@@ -7,9 +7,12 @@ end-to-end tests drive the real curses UI through a pty via expect
 """
 import importlib.util
 import subprocess
+import sys
 import tempfile
 import unittest
 from pathlib import Path
+from types import SimpleNamespace
+from unittest import mock
 
 REPO_ROOT = Path(__file__).resolve().parent.parent
 PICKER = REPO_ROOT / "tools" / "skill_picker.py"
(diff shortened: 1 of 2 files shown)


wanshuiyin:
Reviewed with the usual cross-model pass on top of local checks (`tests/test_skill_picker.py` — 12 passed including your new regression test). Clean on all counts:

- The except-scope is exactly surgical: only `curses.curs_set(0)` is wrapped, only `curses.error` is caught, so `keypad`/drawing/input failures still hit the existing per-group-prompt fallback.
- Your handling matches how Python curses actually behaves — `_curses.curs_set` converts the underlying `ERR` into a raised `curses.error` (with "curs_set() returned ERR" as the message text), which is precisely what the try/except catches.
- The test simulates the failure at the right seam (monkeypatched `curses.curs_set` raising) and asserts the picker still completes rather than falling back — reproduces the Windows Git Bash behavior faithfully.

Nice quality-of-life fix for Windows users. Merging.

PR #383, opened 2026-08-04
feat(idea-discovery): add per-stage evidence gate

3mom3:
## Summary

Adds a deterministic per-stage evidence gate for `/idea-discovery`.

- Records workflow-level `PASS` / `BLOCKED` gate results in the existing `.aris/runs/<run_id>.json` state.
- Adds `tools/idea_discovery_gate.py` to require recorded evidence for `research-lit`, `idea-creator`, `novelty-check`, `research-review`, and `research-refine-pipeline`.
- Writes a visible `BLOCKED: <stage> evidence missing` section into `idea-stage/IDEA_REPORT.md` and exits non-zero when evidence is absent.
- On success, records deterministic acceptance for the validated phases.
- Updates mainline and Codex mirrors, plus the shared resumable-run contract.

## Why

`/idea-discovery` could previously skip a sub-skill yet still emit a complete-looking report. The gate makes missing artifacts and report sections explicit instead of silently proceeding.

## Validation

- `python -m pytest tests/test_run_state.py tests/test_idea_discovery_gate.py tests/test_codex_skill_mirror.py tests/test_skill_groups.py` (52 passed)
- `python -m py_compile tools/run_state.py tools/idea_discovery_gate.py`
- `python tools/check_skills_inventory.py`
- `git diff --check`

## Scope

Focused follow-up for #285. Reuses the existing `run_state.py` state machine rather than adding a parallel workflow-state format.

### DIFF
--- a/skills/idea-discovery/SKILL.md
+++ b/skills/idea-discovery/SKILL.md
@@ -70,7 +70,9 @@ At the end of Phase 5, run:
 ```
 
 The gate writes its result to `gates.idea-discovery-evidence` in the run state.
-On `PASS`, it records deterministic acceptance for each completed phase. On a
+On `PASS`, it records the gate verdict under `gates.idea-discovery-evidence` —
+per-phase acceptance stays with each stage's own cross-model or deterministic
+gate (the evidence gate proves execution, never quality). On a
 non-zero exit, it writes explicit `BLOCKED: <stage> evidence missing` lines to
 the report; do not present the workflow as complete. On `— resume <run_id>`,
 start from the first non-terminal phase and re-run the gate before finalizing.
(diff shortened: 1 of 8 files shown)


wanshuiyin:
Reviewed with the usual cross-model pass on top of local verification. The design is exactly what #285 asked for, and most of it checked out clean on direct inspection: the heading-slug algorithm matches all five SKILL locators (tested concretely), the run_state.py additions are backward-compatible with existing `/research-pipeline` state files (`gates` lazily migrated via `setdefault`, `load_run` read-only, `start_run` untouched for existing files), the mirror adaptation is correct, and the BLOCKED semantics land in both the report and the run state as the issue required.

One doctrine issue survived review, and — nice irony — your own `resumable-runs.md` addition states the correct rule ("a gate does not replace the per-phase acceptance record") while `run()` overreached it: on PASS it called `run_state.accept()` on all five semantic stages, upgrading even `provisional` ones, on the strength of file-exists + heading-present evidence. Per `resumable-runs.md`, a deterministic file-exists check may accept only a *purely mechanical* phase; these five stages carry quality semantics whose acceptance belongs to their own cross-model or stage-level gates — and `resume_point` relies on done-but-not-accepted to know a stage's audit is still pending, so the overreach would have silently removed re-audit on resume.

Since the fix was a deletion (the correct behavior — `record_gate_result()` — was already in place on the line below), I pushed it directly (`ac8d58e`): the accept loop is gone, PASS now asserts phases stay `done`, a new test pins that the gate never calls `run_state.accept`, and the one SKILL sentence claiming "records deterministic acceptance for each completed phase" is reworded on both mirrors. Full suite green.

Thanks for a well-built implementation of the gate — merging.

PR #398, opened 2026-08-11
feat: run Codex literature helpers in isolated uv runtime

2foil:
## Summary
- Add a locked, isolated tools/uv-runtime project and a canonical helper runner.
- Route Codex literature helpers through uv run with the locked project instead of system python3.
- Update the multi-source and standalone literature skills plus integration guidance.

## Validation
- uv lock --check --project tools/uv-runtime
- Wrapper shell syntax check.
- Help-command smoke tests for Semantic Scholar, arXiv, OpenAlex, Exa, and DeepXiv helpers.
- Static check that migrated Skill commands do not execute system python3.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,8 @@ Thumbs.db
 __pycache__/
 *.pyc
 *.pyo
+tools/uv-runtime/.venv/
+tools/uv-runtime/.uv-cache/
 
 # Environment
 .env
--- a/AGENT_GUIDE.md
+++ b/AGENT_GUIDE.md
@@ -167,6 +167,18 @@ Layer 3:  $ARIS_REPO/tools/<helper>                # global fallback
 
 Pick a failure policy from `integration-contract.md` §2 per-helper table: A (gate) / B (side-effect) / C (forensic) / D1 (cascade) / D2 (multi-source aggregate) / E (diagnostic). Each has POSIX-sh + `set -e` + `set -u` safe example blocks.
 
+Python helpers must run through the isolated uv runtime; do not invoke the
+system `python3` directly:
+
+```bash
+ARIS_PY_RUNNER="$ARIS_REPO/tools/uv_run_helper.sh"
+bash "$ARIS_PY_RUNNER" "$HELPER" [args ...]
+```
+
+The runner uses `tools/uv-runtime/pyproject.toml` and `uv.lock` with
+`uv run --project ... --frozen python`. Helper dependencies belong to that
+runtime, not to a downstream research project's environment.
+
 Advisory CI lint at `.github/workflows/lint-skills-helpers.yml` flags hardcoded `python3 tools/foo.py` patterns in PR-modified SKILL.md (warning only, never fails CI). Single-owner helpers (used by exactly one SKILL) live at `skills/<owner>/scripts/<helper>` per Arch C; precedents: `figure-spec`, `paper-illustration-image2`, `experiment-queue`, `render-html`.
 
 ## Cross-Model Protocol
--- a/skills/shared-references/integration-contract.md
+++ b/skills/shared-references/integration-contract.md
@@ -99,6 +99,18 @@ real callers (`AUDIT_VERIFIER`, `TRACE_HELPER`, `WIKI_SCRIPT`,
 `IMAGE2_HELPER`, …) so a single SKILL that resolves multiple helpers
 does not clobber one with another.
 
+Python helpers must additionally resolve the isolated uv runner:
+
+```bash
+ARIS_PY_RUNNER=""
+[ -n "${ARIS_REPO:-}" ] && [ -f "$ARIS_REPO/tools/uv_run_helper.sh" ] && ARIS_PY_RUNNER="$ARIS_REPO/tools/uv_run_helper.sh"
+[ -z "$ARIS_PY_RUNNER" ] && [ -f tools/uv_run_helper.sh ] && ARIS_PY_RUNNER="tools/uv_run_helper.sh"
+```
+
+Invoke Python helpers as `bash "$ARIS_PY_RUNNER" "$HELPER" ...`. The runner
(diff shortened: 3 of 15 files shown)


wanshuiyin (changes requested):
Thanks for this — the problem you are solving is real and we want the change. Running helpers under the ambient `python3` means `requests`, `exa-py` and `deepxiv-sdk` have to exist in whatever environment the user's *research project* happens to have, and a locked runtime is the right way to decouple that. The wrapper itself is well written: quoting, argument forwarding, runtime-relative lookup and the `UV_CACHE_DIR` preservation are all sound, and the 530-line `uv.lock` is the mechanism that makes the runtime reproducible, not bloat.

Requesting changes on four things, all of which are supported install paths that stop working on merge. This was reviewed by a second model as well as by hand; the findings below are ones I verified against the branch myself.

---

### 1. `uv` becomes a hard requirement that nothing installs

`tools/uv_run_helper.sh:14-17` exits 127 when `uv` is absent, and the new contract text says system `python3` "is not an allowed fallback". But `uv` is currently referenced only as an optional alternative in the Vast/Zotero setup — it is in neither `tools/install_aris.sh` nor `tools/install_aris_codex.sh`, not in `README.md`'s prerequisites, and not in `skills/skills-codex/README.md`'s installation section. This PR touches no install path.

I checked this repo's own machine: no `uv` on `PATH`, so `/exa-search`, `/openalex` and `/deepxiv` stop working on the next run, and `/semantic-scholar` hits `skills/skills-codex/semantic-scholar/SKILL.md:84-87` — `refusing bare python3 fallback` then `exit 1`.

To be fair to the PR: this is not "everything breaks". `/research-lit` and `/arxiv` still degrade to web search and keep going. But three skills stop outright and one hard-exits, on a machine that was working before `git pull`.

**This project's rule is that a tightening ships its migration guidance in the same commit.** So please add, here rather than in a follow-up: `uv` preflight/install in `tools/install_aris_codex.sh`, the prerequisite in `README.md` and `skills/skills-codex/README.md`, and an upgrade note for existing users. Note the installer alone is not enough — `README.md:1519-1523` tells symlink users that ordinary content updates arrive via `git pull` and that rerunning the installer is mainly for added or removed skills, so someone who only pulls never runs your new preflight. The wrapper's error message should therefore name the exact install command.

**On the shape of the fix:** I originally wanted a `python3` fallback with a warning, and I have been argued out of it. A fallback would recreate the drifting-dependency behaviour the PR exists to remove and would leave two materially different execution paths, which is worse than a clear failure. Hard prerequisite plus real guidance is the right call — the guidance is the part that is missing.

### 2. The mainline contract now forbids what the mainline skills do

The PR edits `skills/shared-references/integration-contract.md` — the **mainline** contract — to forbid a `python3` fallback, and adds the same blanket paragraph to `AGENT_GUIDE.md`. It changes zero mainline `SKILL.md` files.

Those mainline skills reference that exact document by relative path (`skills/arxiv/SKILL.md`, `skills/deepxiv/SKILL.md`, `skills/exa-search/SKILL.md`, `skills/openalex/SKILL.md`, `skills/semantic-scholar/SKILL.md`, `skills/research-lit/SKILL.md` all link `../shared-references/integration-contract.md`), and they still contain direct `python3` invocations — 5, 5, 6, 5, 5 and 14 of them respectively, with zero references to the new runner.

So on merge, six mainline skills violate the mainline contract that the same commit wrote. For a Codex-scoped PR the clean fix is to revert the mainline contract edit and the `AGENT_GUIDE.md` paragraph, keeping the change in `skills/skills-codex/shared-references/`. Migrating the Claude side is a coherent separate PR and we would welcome it.

### 3. Copied Codex installs cannot resolve the runner at all

Copied installs are a documented, supported path (`skills/skills-codex/README.md:68-91`), and `tools/smart_update_codex.sh:129` writes `$HOME/.aris/repo` precisely so copied skills can still find `tools/`.

The new resolver blocks do not use it. Every rewritten skill and `skills/skills-codex/shared-references/integration-contract.md` resolve only through `.aris/installed-skills-codex.txt` and a relative `tools/uv_run_helper.sh`. A copied install has neither — the copied directory holds `SKILL.md` and nothing else. So even a user who installs `uv` correctly still lands on the `exit 1` in §1.

`skills/skills-codex/shared-references/wiki-helper-resolution.md:12-21` already has the pointer-read pattern for exactly this; please reuse it in the contract and in each rewritten resolver chain.

### 4. Native Windows Codex installs have no `bash`

`README.md:1537-1539` advertises PowerShell Codex installation with no WSL required, and `tools/install_aris.ps1:70` accepts `-Platform codex`. The PR ships `tools/uv_run_helper.sh` only, and the rewritten skills invoke it as `bash "$ARIS_PY_RUNNER" ...` in 36 places. Native Windows does not guarantee `bash` on `PATH`, and installing Windows `uv` does not help. This needs a PowerShell entrypoint and the matching prerequisite in `install_aris.ps1`.

---

### Also worth fixing

**`/research-lit` still bypasses the isolated runtime.** `skills/skills-codex/research-lit/SKILL.md:255-260` keeps an `elif command -v deepxiv` branch that shells out to the ambient `deepxiv` CLI. That is the same class of drift the PR removes elsewhere, and it is a stronger bypass than the `python3` one the contract now forbids. Either drop the branch and degrade the optional source, or say explicitly in the contract that ambient first-party CLIs are exempt — but the current state contradicts itself.

**One line on first use.** `uv run --frozen` does not update the lock but still materialises the environment on first run, so the first invocation needs network, and may need to fetch a Python in the declared `>=3.11,<3.15` range. Worth a sentence in `tools/uv-runtime/README.md` so an offline user knows why it stalled.

---

None of the above touches the design. Isolated locked runtime: yes. What is missing is the install path that makes it reachable, the two supported install shapes it currently cannot serve, and keeping the two skill trees consistent with the contracts they point at. Happy to review again as soon as you push.

PR #401, opened 2026-08-14
fix(aris-cli): honor Windows and macOS system proxies

JasmineLCY:
## Summary

- explicitly enable the reqwest `system-proxy` feature for the final ARIS CLI dependency graph
- restore Windows and macOS desktop proxy discovery that was disabled together with reqwest default features
- update the lockfile with the platform-specific proxy dependencies

## Why

ARIS intentionally disables reqwest default features in order to select `native-tls`. In reqwest 0.12.28, that also disables `system-proxy`, so users who configure a proxy only through Windows or macOS settings can bypass that proxy and hit a direct-connect timeout.

This keeps the existing TLS choice while enabling only the missing proxy feature. The CLI owns the Gemini executor, and Cargo feature unification enables it for the shared reqwest version in the final binary.

## Validation

- `cargo build --workspace --all-targets`
- `cargo test --workspace -- --test-threads=1` — 559 passed, 1 ignored
- `cargo tree -p aris-cli -e features -i reqwest` confirms `reqwest feature "system-proxy"`
- a focused `x86_64-pc-windows-msvc` cargo check using the same pinned reqwest 0.12.28 feature set passes, including `windows-registry`

A Windows system-proxy runtime smoke test was not available locally. A full workspace cross-check from macOS reaches the existing `onig_sys` C dependency and requires Windows SDK headers, so the repository Windows build job remains the authoritative platform compile gate.

Related to #163

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -145,6 +145,16 @@ dependencies = [
 
 [[package]]
 name = "core-foundation"
+version = "0.9.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "91e195e091a93c46f7102ec7818a2aa394e1e1771c3ab4825963fa03e45afb8f"
+dependencies = [
+ "core-foundation-sys",
+ "libc",
+]
+
+[[package]]
+name = "core-foundation"
 version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


JasmineLCY:
CI is currently marked `action_required` because GitHub requires maintainer approval for this first-time fork workflow. Once it is approved, I will monitor the cross-platform jobs and address any failures.

Issue #87, opened 2026-03-27
Security: Research output injected into Claude context without sanitization


manja316:
Hi,

While reviewing Claude automation tools for security, I noticed a prompt injection surface in ARIS worth flagging.

### Finding

| # | Issue | Severity | Location |
|---|-------|----------|----------|
| 1 | Web research results fed to Claude without output sanitization | MEDIUM | Research output → Claude prompt pipeline |
| 2 | No adversarial content detection in fetched web pages | MEDIUM | Web fetch → markdown → context |

**Trust Score: 55/100** (C+ — powerful automation with minimal input filtering)

### Details

ARIS fetches web content during research and feeds it into Claude's context as part of the research pipeline. Web pages can contain adversarial prompt injection payloads (hidden text, CSS-invisible divs, markdown injection). Without sanitization, these become part of Claude's instruction context.

**Attack scenario:** A researcher targets a topic where an attacker has placed prompt injection payloads on indexed web pages. When ARIS fetches and processes these pages, the injected instructions modify Claude's behavior — potentially exfiltrating research data, altering conclusions, or executing unintended actions.

**Impact:** Research integrity compromise. Claude's output could be manipulated by adversarial web content.

**Suggested fix:** Strip or escape known prompt injection patterns from fetched web content before including in Claude's context. Consider a content sanitization layer between web fetch and LLM prompt construction.

---

Part of MCP security research at [LuciferForge](https://github.com/LuciferForge).

PR #408, opened 2026-08-18
fix(idea-creator): scan wiki context before reading

JasmineLCY:
## Summary

- scan cached query packs with `threat_scan.py --scope strict` immediately before reading them
- scan stale or missing packs immediately after the normal one-time rebuild
- on a threat hit, scanner error, or unresolved scanner, leave the raw pack untouched and continue primary ideation without wiki context
- keep the main and Codex skill mirrors plus injection-hygiene docs in sync

## Scope

This addresses the persistent query-pack reinjection path described in #87. Raw WebSearch/WebFetch results and fetched abstracts remain explicitly out of scope; this PR does not claim to sanitize the entire web-research surface.

Refs #87

## Testing

- `uv run --with pytest pytest -q tests/test_idea_creator_query_pack_scan.py` (6 passed)
- `uv run --with pytest --with httpx pytest -q` (638 passed, 17 skipped, 4 subtests passed)
- `bash tools/lint_skills_helpers.sh` (advisory only; existing documentation example)
- `git diff --check`

### DIFF
--- a/skills/idea-creator/SKILL.md
+++ b/skills/idea-creator/SKILL.md
@@ -92,12 +92,11 @@ THREAT_SCANNER=".aris/tools/threat_scan.py"
 [ -f "$THREAT_SCANNER" ] || { [ -n "${ARIS_REPO:-}" ] && THREAT_SCANNER="$ARIS_REPO/tools/threat_scan.py"; }
 [ -f "$THREAT_SCANNER" ] || THREAT_SCANNER=""
 
-# ARIS_QUERY_PACK_SAFE_VIEW_START -- exercised by
+# ARIS_QUERY_PACK_SCAN_START -- exercised by
 # tests/test_idea_creator_query_pack_scan.py; keep both skill mirrors identical.
-aris_prepare_query_pack_view() {
+aris_scan_query_pack() {
   local query_pack_raw="$1"
-  local query_pack_candidate query_pack_scan_status
-  QUERY_PACK_SAFE_VIEW=""
+  local query_pack_scan_status
   QUERY_PACK_SCAN_RESULT="error"
 
   if [ -z "${THREAT_SCANNER:-}" ] || [ ! -f "$THREAT_SCANNER" ]; then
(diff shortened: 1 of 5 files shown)


JasmineLCY:
CI is currently waiting for maintainer approval to run workflows from this fork. Local validation is complete: targeted tests passed (6/6) and the full suite passed (636 passed, 17 skipped, 4 subtests passed).

JasmineLCY:
Pushed follow-up `9ab00bf`: Phase-0 resolver paths now guard an unset `HOME` under `set -u` while preserving pointer/global fallback when `HOME` exists; regression coverage was added. Targeted and full suites pass locally.

JasmineLCY:
Rebased onto current `origin/main` (`f4f20f9`) with both #87 commits preserved byte-for-byte by `git range-diff`; new head is `518cf2e`. Revalidation: targeted 6 passed, full suite 638 passed / 17 skipped / 4 subtests passed, and diff-check is clean.

wanshuiyin:
Reviewed (maintainer + second model). The injection path is real and reachable — the wiki ingests WebSearch/WebFetch-derived material, `query_pack.md` gets rebuilt from it, and `/idea-creator` consumes the pack as prompt context, so a cached pack bypassing rebuild-time scanning is a genuine gap. Wiring the pre-existing `threat_scan.py` into the read path is the right fix, and the fallback you chose — continue primary ideation without wiki context rather than hard-fail — respects installs that cannot resolve the scanner (raw copied installs without the updater pointer). Good.

One reduction requested before merge: **the private permission-sealed snapshot + quarantine-copy + rebuild + rescan sequence is more machinery than this threat needs.** ARIS assumes a cooperating operator on their own machine; the TOCTOU window between scan and read is not adversarially exploitable in that model, and each extra moving part here is something every future maintainer must keep in sync. Scan immediately before reading; on a hit or scanner error, keep the raw pack where it is and skip wiki context for this run. That preserves your fail-closed semantics and the evidence, with a fraction of the mechanism.

Keep: scan-on-read, strict scope, the no-wiki-context degradation, mirror/doc sync. Also appreciated the honest scope note that raw WebSearch surfaces remain out of scope.

JasmineLCY:
Addressed in `29d813c`. The read gate now runs the strict scanner directly against `query_pack.md` immediately before Read. A hit, scanner error, or unresolved scanner leaves the raw pack untouched and skips wiki context for that run; the snapshot, permission sealing, quarantine copy, and rejected-pack rebuild/rescan path are removed. Stale/missing packs retain only the normal one-time rebuild before the same scan-on-read gate. Both skill mirrors, hygiene docs, tests, PR title, and PR body are updated. Validation: targeted 6 passed; full suite 638 passed / 17 skipped / 4 subtests passed; helper lint advisory only; diff-check clean.

wanshuiyin:
Merged. Verified on the new head that scan-on-read survived and the snapshot/quarantine machinery is gone — the "Do not copy, quarantine, rebuild, rescan" line even forecloses it explicitly, which is more than we asked. Thanks.

PR #409, opened 2026-08-18
fix(idea-discovery): require durable reviewer evidence

### LINKED ISSUE #375: Bug! The workflow seems to have been disrupted by the updated Claude.
Now that the process has been eliminated, the number of times Claude calls upon Codex during execution has decreased significantly (the process requiring Codex review was not carried out—it was omitted), and the level of cognitive effort has also decreased.

JasmineLCY:
## Summary

- require durable, identity-bearing review receipts for `novelty-check` and `research-review` before the idea-discovery evidence gate can pass
- validate recorded reviewer provenance and a plain non-empty first-match report section
- document honest Claude/Codex receipt paths: only positive verdicts may call `accept` or `mark-provisional`; negative verdicts remain `done` and keep the workflow `BLOCKED`

## Tests

- `uv run --with pytest pytest -q tests/test_idea_discovery_gate.py` (21 passed)
- `uv run --with pytest --with "httpx>=0.27,<1.0" pytest -q` (647 passed, 17 skipped, 4 subtests passed)
- `git diff --check origin/main`

Fixes #375

### DIFF
--- a/skills/idea-discovery/SKILL.md
+++ b/skills/idea-discovery/SKILL.md
@@ -45,7 +45,8 @@ helper chain used by `/research-pipeline`: `.aris/tools/` → `tools/` →
 the final report is `BLOCKED`; do not silently continue without a state record.
 
 For a new run, derive `<run_id>` from the direction slug and date, then start
-this ordered state record:
+this ordered state record with `--executor <actual-Claude-model>` (for example,
+`claude-sonnet-4.5`):
 
 ```text
 research-lit,idea-creator,novelty-check,research-review,research-refine-pipeline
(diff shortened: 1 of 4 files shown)


JasmineLCY:
CI is currently marked `action_required` because GitHub requires maintainer approval for this fork workflow. Once approved, the test, inventory, and helper-resolution workflows can run.

wanshuiyin:
Reviewed (maintainer + second model). The core of this is right and we want it: #375 was Claude simply omitting the Codex review after a harness change, and requiring the two reviewer-bearing phases to carry a real reviewer model + review-session handle before the gate passes is exactly proportionate to that failure. The upgrade story is also acceptable — an in-flight run whose phases are merely `done` re-records or reruns the review, which is recoverable and is the point of the fix. We will not ask for a compatibility exemption; that would preserve the omission.

One trim requested before merge, and it is a trim, not a rethink: **drop the adversarial Markdown handling** — the fenced-code pseudo-heading exclusion, HTML-comment stripping, duplicate-heading logic (~half the new gate code). A report whose first-match section is empty is worth catching; a cooperating operator hiding a heading inside a code fence is not a case that occurs here, and parsing machinery for it is exactly the fortification this repo screens against. A plain non-empty-section check plus your receipt/provenance validation is the whole fix.

Keep: the receipt requirement, provenance validation, positive-verdict-only accept/mark-provisional, all of it. CI is green on your branch.

JasmineLCY:
Implemented the requested trim in 55366b5: section validation is now a plain first-match/non-empty-body scan, and the fenced-code, HTML-comment, and duplicate-heading-specific parsing/tests are removed. Receipt/provenance validation and positive-verdict-only `accept` / `mark-provisional` behavior remain intact.

Verified with the targeted gate tests (21 passed), the full suite (647 passed, 17 skipped, 4 subtests passed), and `git diff --check origin/main`.

wanshuiyin:
Merged. Trim verified against the new head before merging — zero fenced/HTML-comment handling left in the gate, receipts and provenance intact, 647 green. Thanks for the fast, exact turnaround; both this and #408 were implemented precisely to the review with no over-correction, which is rarer than it should be.

PR #411, opened 2026-08-18
feat(llm-chat): add safe Atlas Cloud configuration

binyangzhu000-sudo:
## Summary

- add Atlas Cloud as an OpenAI-compatible `llm-chat` reviewer configuration
- add `LLM_RETRY_ON_504` so paid POST calls can opt out of automatic 504 retries
- preserve the existing retry and fallback behavior by default

## Testing

- `python3 -m unittest tests.test_llm_chat_server -v` (25 passed)
- `python3 -m unittest discover -s tests -v` (273 passed, 4 skipped, 2 pre-existing Python 3.9 import errors; reproduced on the unmodified base)
- `python3 -m py_compile mcp-servers/llm-chat/server.py tests/_llm_chat_helpers.py`
- `git diff --check`

No live generation request was sent. The Atlas model catalog was read to confirm `deepseek-ai/deepseek-v4-pro` supports `openai.chat.completions`.

### DIFF
--- a/docs/LLM_API_MIX_MATCH_GUIDE.md
+++ b/docs/LLM_API_MIX_MATCH_GUIDE.md
@@ -108,10 +108,34 @@
 
 ### 常用审查器提供商
 
-| 提供商 | LLM_BASE_URL | LLM_MODEL |
-|--------|--------------|-----------|
-| DeepSeek | `https://api.deepseek.com/v1` | `deepseek-chat` |
-| MiniMax | `https://api.minimax.io/v1` | `MiniMax-M3` |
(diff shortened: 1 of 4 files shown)


wanshuiyin:
Closing. The retry knob is mislabeled — `max_attempts = 3 if RETRY_ON_504 else 1` disables retries and the fallback-model attempt for *every* failure class, not just 504, so the env var does not do what its name and docs claim. And the substantive content of the PR is a vendor placement: a provider row, endpoint URL and model id for a service unaffiliated with this project.

If the double-billing concern on gateway timeouts is real for you in practice, a scoped fix — skip the retry only when the response status is 504 and the request is verdict-bearing — with a test that locks that behavior would be reviewed on its merits.

PR #414, opened 2026-08-19
fix(readme): restore the broken Star History chart

Dessalines39394:
The Star History chart in both README.md and README_CN.md no longer renders. Switching the chart image and its link to the maintained provider so the chart displays again. The GitHub stars badge is left on its original provider.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1749,7 +1749,7 @@ If you use ARIS in your research, please cite:
 
 ![GitHub stars](https://img.shields.io/github/stars/wanshuiyin/Auto-claude-code-research-in-sleep?style=social)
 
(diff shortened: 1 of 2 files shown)


wanshuiyin:
Thanks for the PR, but the premise does not hold: `api.star-history.com` is serving this exact chart right now — HTTP 200, `image/svg+xml`, 60 KB (checked today from a clean network). The chart is not broken.

`star-history.dera.page` is not the maintained provider — star-history.com is the project's own domain, and dera.page is an unaffiliated mirror. Pointing the README of a 13k-star repo at a third-party mirror hands control of that image (and the click-through) to whoever operates the mirror, for no benefit while the official endpoint works.

If the chart genuinely stops rendering for you, an issue with a screenshot and the failing URL would be welcome. Closing.

PR #434, opened 2026-09-08
Add OrcaReplay to External Projects & Docs

xizhuomengcontin:
Adds [OrcaReplay](https://github.com/Continuum-AI-Corp/OrcaReplay) to **🌐 External Projects & Docs**.

**Disclosure: I work on it.** Apache-2.0, free, no paid tier.

## Why it is relevant to ARIS specifically

ARIS runs an autonomous loop overnight and you read the result in the morning. When that loop goes somewhere wrong, the interesting question is *where* — and re-running it to find out costs another night and another pile of tokens, with no guarantee the same thing happens twice.

OrcaReplay records the run while it happens, below the harness. No SDK, no exporter, no plugin: it moves a base-URL variable for that process, or terminates TLS for one named host. The trace keeps the **verbatim request and response bytes**, plus tool calls and per-turn file changes, on one timeline. Two things follow:

- `orca replay <run>` re-executes the run **with the network off** (`egress=blocked`), served the recorded bytes back. Free, deterministic, as many times as you like.
- `orca replay <run> --from N --model <other>` restores the git tree to that checkpoint in an isolated worktree, replays up to it, then goes **live** from there on a different model. `--verify "<cmd>"` makes the exit code the verdict — which is the cross-model review loop ARIS already does, but scored rather than read.

It works with the executors ARIS targets: Claude Code, Codex CLI and OpenClaw all have adapters, and so does goose.

## Verified, not asserted

Claude Code is [walked through against a real bug fix](https://github.com/Continuum-AI-Corp/OrcaReplay/blob/main/docs/validation.md). goose 1.49.0 was driven end to end three times today — `1 failed` → agent fixes it → `1 passed`, then strict offline replay `reused=8/8 exact=8 divergences=0` every time. Hermes v0.21.0 records and replays too.

`npm i -g orcareplay` · ~1,700 tests · trace-format conformance suite · listed in the official MCP Registry.

## One honest limit

Forking onto a different model works for harnesses redirected by a base-URL variable. It does **not** currently work for a harness captured by terminating TLS — the `--model` override is accepted and has no effect there. That is [an open bug on our side](https://github.com/Continuum-AI-Corp/OrcaReplay/issues/49), not a footnote I would rather you discovered later.

Happy to shorten the row or move it if `Reproducibility / DevOps` is not the domain label you would pick.


pikadothub:
Holy s.. wtf i dif, I saw It now

wanshuiyin:
Merged, thanks — and thanks for the disclosure and the honest note on #49. I trimmed the row to the length of the other entries, added the Chinese-README row, and added one sentence that the trace stores full request/response bytes (prompts and keys included), since most ARIS users feed unpublished work through the agent.

PR #437, opened 2026-09-14
feat: package the skill set as an installable plugin

white-drizzle:
## What

Packages ARIS as an installable plugin, so the skill set can be added through the plugin flow as an alternative to the existing installers:

```
claude plugin marketplace add wanshuiyin/Auto-claude-code-research-in-sleep
claude plugin install aris@aris
```

Two new files, plus a short note in both READMEs (10.2 Install Skills):

- `.claude-plugin/marketplace.json` — declares marketplace `aris` with one plugin, `aris` (source `./`)
- `.claude-plugin/plugin.json` — plugin metadata; skills are picked up from the existing `skills/` directory by convention, so **no skill file is moved, renamed or duplicated**, and the Codex mirror is untouched
- `README.md` / `README_CN.md` — one paragraph documenting the route and its trade-off

## Verification

Claude Code 2.1.220; marketplace added from a local directory, plugin installed and inspected:

- `claude plugin validate .claude-plugin/marketplace.json` → passed, no warnings
- `claude plugin details aris` → **Skills (82)**, Agents (0), Hooks (0), MCP servers (0), LSP servers (0); projected always-on cost ~5.7k tokens
- all 82 mainline skills resolve, including the ones added since the docs snapshot
- `tools/check_skills_inventory.py` → `ARIS skill inventory is consistent.`
- the test install was uninstalled afterwards; installing modifies nothing inside the repository

## Scope and trade-offs

- The plugin installs **all 82 skills at once** — it complements rather than replaces `tools/install_aris*.sh`, which keep group selection (`--groups` / `--skills` / `--exclude`), the `.aris/installed-skills.txt` reconcile manifest, and the Codex mirror installs.
- Payload is currently the repository root. If a skills-only payload is preferred, the marketplace entry can switch to a `git-subdir` source pointing at `skills/` (that would need a small manifest inside `skills/`) — happy to change it.
- `mcp-servers/` and `tools/` are intentionally not part of the plugin yet: bundling the `codex-exec` MCP server would make the plugin auto-connect it, which felt like a separate decision to make.
- The Codex / Gemini review overlays (`skills/skills-codex*`) are not exposed as separate plugin entries yet; they can be added as extra entries if you want them installable too.

Opened after the roadmap note in #96.


wanshuiyin:
Merged, thank you — the manifests were exactly right and I built the rest on top in a follow-up commit rather than sending you back:

- `/aris:setup` (`commands/setup.md`): checks codex is installed and logged in, registers the Codex reviewer bridge that ships inside the plugin (`mcp-servers/codex-exec/server.py`, ARIS's stand-in for the `codex mcp-server` that codex-cli 0.154 removed) under the `codex` server name the skills hardcode, and points `~/.aris/repo` at the plugin so helper scripts resolve. Deliberately not a bundled `.mcp.json`: a plugin-provided server would surface as `mcp__plugin_aris_codex__codex`, and 47 skills say `mcp__codex__codex`.
- Verified with `--plugin-dir`: 84 skills appear as `aris:*`, sibling skills still call each other by short name inside the namespace, and `/aris:setup` works both on a machine with nothing registered and on one already set up.
- Dropped the pinned `version` so plugin users receive updates as the repo moves, and made the descriptions count-free.
- README (EN/CN) now describes the route as install → `/aris:setup` → restart, with the `aris:` prefix.

The plugin route is live in the README's install section and the News.

PR #438, opened 2026-09-14
feat(skill): /research-implement-feature — build from "implement X" with a declared assumption ledger

heroarmor:
## What this adds

`/research-implement-feature` — a build skill for the one request shape ARIS
doesn't currently have an entry point for: **"just implement X for me"**, where
the user has a capability in mind rather than an experiment plan, and doesn't
want to be interviewed about it first.

It resolves that request the way I think is the honest one: **stay autonomous,
stop being silent.**

## Why not a mode on `/research-pipeline`

Fair question, and I went back and forth on it — ARIS already has a lot of top-level
entry points, and another one is a real cost.

The reason it ended up separate is the **input contract**, not the workflow shape:

- `/research-pipeline` answers *"what should we research?"* — it decides the
  research question for you, which is exactly what you want from it.
- This skill answers *"build the thing I already decided on"* and decides
  **nothing** without writing it down.

Folding one into the other would mean a mode flag that changes what the skill is
allowed to decide on your behalf, which reads like two skills sharing a name.

They compose rather than compete: a pipeline run can delegate its build stage
here and inherit the assumption ledger. There's a routing table and a
`Relationship to /research-pipeline` section in the SKILL so users landing on
either one get pointed at the right entry point. Happy to reshape this if you'd
rather it live as a pipeline mode.

## The two mechanisms

**1. Spine before features.** F0 is a walking skeleton — the thinnest path from a
real entry point to a real artifact, placeholders inside — and it must exit 0
before any feature is added. Then one rung at a time, each with ONE runnable
acceptance command, each requiring every earlier rung still green. A rung that
exhausts its fix budget **halts the ladder** rather than getting reordered to the
end so the progress bar looks fuller.

**2. Declare before you act.** Every decision the request left under-determined
gets a ledger row *before* the code that depends on it exists, class-tagged:

| class | means |
|---|---|
| `cosmetic` | naming, layout, log format |
| `local` | one module's internals |
| `interface` | changes call sites, configs, artifact schemas |
| `semantic` | **changes what a result would MEAN** — metric definition, eval split, normalization, what counts as a baseline |

`semantic` rows get their own block at the top of the final report and are never
collapsed into a count. Each row carries its **reversal cost** and an override
string, so an assumption you dislike is one glance from being fixed.

Code sites carry `ASSUMPTION[A-001]` markers, and a deterministic `grep` + `comm`
bijection check halts a rung if any marker lacks a ledger row. That check is
shell, not a model's opinion — "I believe the ledger is complete" is never
accepted in its place.

## The part I'd most like reviewed: Phase 4

The ledger only records what the implementer *noticed* assuming. So Phase 4 sends
the spec, the ledger, and the **raw diff** — no Claude-written summary, per
`reviewer-independence.md` — to a reviewer from a different model family, with
one question:

> what does the code decide that the request didn't determine and the ledger
> doesn't declare?

Verdict lands verbatim in `implement-stage/SILENT_ASSUMPTION_SWEEP.json`. At
`assurance: submission`, undeclared **semantic** assumptions block the final
report.

This is the one gate a same-family pass structurally cannot do — you can't audit
yourself for what you stopped noticing. It's also the direct answer to the
failure mode that motivated the skill: an implementation that quietly decides
what the research will later claim.

## Interaction modes

`— ask: never | semantic | all`, one axis. The ledger row is the unit of
ambiguity, so "which ambiguities get asked about" and "which rows get written"
are the same set — `ASK` just picks which classes are put to the author instead
of decided for them.

| mode | asks about | blocking |
|---|---|---|
| `never` *(default)* | nothing — declare and proceed | no |
| `semantic` | `semantic` rows only | at batch points |
| `all` | every ambiguity, any class (**confirm mode**) | at batch points |

Default is `never` so the skill stays usable unattended. `all` is for when you're
at the keyboard and want the build to match what's in your head.

Confirm mode is a **batch gate, not a dialogue**: questions collected and asked
together, at most 4 per call, ordered by blast radius, cheap classes bundled into
one question, your default always option 1 so accepting everything is one
keystroke and yields exactly what `ask: never` would have. Batch points are B0
(end of Phase 0), B*i* (start of each rung), and Bd (a debugging fork that is
itself an ambiguity — asked before the fix, since a semantic choice buried in a
bug fix is the easiest one to lose).

Two things it refuses: silently downgrading `ask: all` to `never` when nobody
answers (it says the run is unattended and stops), and treating a `user`-sourced
row as exempt from Phase 4 (an answer makes a row *declared*, not *correct*).

## Gate provenance

Per `acceptance-gate.md`, every stop gate is classified in the SKILL. The build
loop is **Type-A throughout** — exit codes, counters, `grep`. On a green run it
says *"the spine runs and every MUST rung's check passed"* and never that the
implementation is correct or that a number means anything; that verdict stays
with `/experiment-audit` and `/result-to-claim`.

The one Type-B gate it owns is the Phase 4 sweep, routed cross-family per
`reviewer-routing.md` (regular tier, `gpt-6-astra` + `xhigh`, both pinned on the
first call).

## Codex mirror

`skills/skills-codex/research-implement-feature/SKILL.md` carries the honest
degradation: in the base mirror executor and reviewer are both GPT, so the sweep
records `review_independence: same-family`, `acceptance_status: provisional`, and
is documented as able to **flag but never acquit** — at `assurance: submission` a
mirror `verdict: clean` doesn't by itself clear the gate. Deterministic checks
are unaffected, since a process is not a model family.

## Inventory / CI

Registered per the usual set: mirror, `docs/SKILLS_CATALOG.md`,
`tools/skill-groups.tsv`, the live counts in README / README_CN / AGENT_GUIDE /
ARIS_INTRO (`.md` + `.html`, heading + anchor + TOC link) / both skills-codex
READMEs, and the assertion in `tests/test_codex_skill_mirror.py`. Historical
changelog and news entries mentioning 82 are left as written.

**I put it in the `experiments` group and under "🧪 Experiments & Infrastructure"
in the catalog — move it wherever you'd rather have it, that placement is a guess.**

## Testing

```
python tools/check_skills_inventory.py      → ARIS skill inventory is consistent.
python -m pytest tests/ -q                  → 725 passed, 21 skipped
bash tools/lint_skills_helpers.sh           → no new findings (the skill invokes no tools/ helper)
```

Verified against a clean worktree of `f1bd907` with the same dependency set CI
installs (`pytest httpx`): **zero new failures, zero newly-skipped tests.**

The bijection check in the SKILL was run against a fixture with one orphan marker
and one ledger row with no code site — it flags the orphan and tolerates the
code-site-free row, as documented.

## Open questions for you

1. Group / catalog placement — `experiments` is my guess.
2. Phase 4 runs at the **regular** reviewer tier (`xhigh`). It's a one-shot
   verdict-bearing audit, which is what `ultra` is for, but the deep-audit row in
   `reviewer-routing.md` enumerates specific skills and I didn't want to add
   myself to it unilaterally. Say the word and I'll move it.
3. `USER_GUIDE.md` alongside the SKILL is a departure from the one-file norm —
   the SKILL is the model-facing contract and the guide is human-facing. Happy to
   fold it into the SKILL or drop it if you'd rather keep skill dirs single-file.


wanshuiyin:
Thank you for this — "just implement X, no experiment plan" is a request shape ARIS genuinely had no entry point for, and the core here is exactly right: spine first, one runnable check per rung, assumptions written before the code that depends on them, a different model family reading the raw diff for what went undeclared, and "the checks passed" kept separate from "the method works". We'd like to bring it in, and we'd like it to sit near the front of the catalog.

Could we simplify the bookkeeping around that core while keeping the assumption ledger and the independent sweep? Concretely:

1. **Ledger columns** — ID · what was under-determined · chosen · class · source carries the idea. Rejected alternatives, reversal cost and the override string are good prose where a decision is genuinely contested, but as required columns every row becomes a form. And only `interface` / `semantic` decisions really need rows; a naming or module-internal choice can just be made.
2. **`ASSUMPTION[A-001]` markers + the grep/comm check** — it confirms markers have rows, not that decisions have rows, and the completeness question is the one Phase 4 already owns. We'd drop the markers and the check.
3. **Placeholders** — "label every stub, never let a stub number reach a results path" is the rule worth keeping; numbered `P-IDs` and the repo-wide grep gate can go.
4. **Artifacts** — `SPEC.md`, `ASSUMPTIONS.md` and the sweep JSON carry the skill; the ladder and build log could be one build note.
5. **`ask:` modes** — `never` (default) and `semantic` cover the real cases; `all` with its batching rules adds a lot of interaction overhead for this workflow.

A few concrete fixes:

- Phase 4 prompt, in both the mainline SKILL and the Codex mirror: please include the `review-scope-limits.md` block (`skills/shared-references/`) — every reviewer prompt in ARIS carries it. The audit only reads, so `sandbox: read-only`; and `approval-policy` can go — the `codex` MCP server is now ARIS's own bridge over `codex exec` (`mcp-servers/codex-exec/`) and doesn't take that argument.
- `git diff <base>` — `<base>` is never established. Recording the starting commit in `SPEC.md` at Phase 0 would let the reviewer get the actual diff.
- A "you decide" answer lands as `Source: default`, but the sweep reads `default` rows in a covered class as ambiguities the skill never recognised — the two collide. The row already carries `deferred_to_author: true`; the sweep prompt could key on that instead.
- The Codex-mirror `SKILL.md` points at a claude-review overlay for this skill; there isn't one (the overlay generator keeps an explicit list).
- `USER_GUIDE.md` has already drifted from the SKILL (the append/resume behaviour it promises, the install line without shared references). The worked example is good — folded into the end of `SKILL.md`, it can't drift.

If it's easier, we're glad to do the trim on top of your branch and send it back for your look — just say which you'd prefer. Thanks again for putting this much care into it.

heroarmor:
Thanks — the trim is pushed (`134b4c5`). All five simplifications and all five fixes are in; details below, including the one place I went a step further than you asked so you can push back on it.

### Trims

1. **Ledger columns** — now `ID · under-determined by the request · chosen · class · source`. Rejected alternatives, reversal cost and the override string moved to a `## Notes` section under the table, written only where a decision is genuinely contested. And only `interface` / `semantic` decisions get rows: naming, log format, file layout and module-internal choices are just made. `cosmetic` and `local` are gone as ledger classes — the SKILL now says explicitly that a ledger which logs variable names buries the two rows that decide what the work will later claim.
2. **Markers + bijection check** — removed. Gone from the acceptance-gate table, Phase 3, and the mirror. Both files drop the `grep`/`comm` block entirely.
3. **Placeholders** — kept "label every stub at its site, a stub number never reaches a results path, a rung isn't green while a stub it was meant to retire is live." Numbered `P-IDs` and the repo-wide grep gate are gone; the marker is now a plain `# PLACEHOLDER: …` comment.
4. **Artifacts** — one `BUILD_NOTE.md` replaces `FEATURE_LADDER.md` + `BUILD_LOG.md`. **I also folded `DEFERRED.md` and `BLOCKERS.md` into it** as `## Deferred` / `## Blockers` sections, since both are content about the same ladder and your list of what carries the skill read as the full artifact set. That leaves four files: `SPEC.md`, `ASSUMPTIONS.md`, `BUILD_NOTE.md`, `SILENT_ASSUMPTION_SWEEP.json`. Say the word if you'd rather keep those two as separate conditional files.
5. **`ask:` modes** — `never` (default) and `semantic` only. `all` and the seven batching rules are gone. What survives is one short paragraph: collect the batch, ask in one call, default as option 1 so accepting is one keystroke and yields exactly what `ask: never` would, "you decide" falls back and is never re-asked, empty batch is silent, don't pair with an unattended cadence.

### Fixes

- **`review-scope-limits.md` block** — inlined verbatim in the Phase 4 prompt in both the mainline SKILL and the Codex mirror, matching how `/experiment-bridge` and `/research-review` carry it. Added to both *See Also* lists.
- **`sandbox` / `approval-policy`** — the mainline Codex call is now `{"model": "gpt-6-astra", "config": {"model_reasoning_effort": "xhigh"}, "sandbox": "read-only", "cwd": "<repo root>"}`. `approval-policy` is gone. The mirror's `spawn_agent:` block carries only `model` / `reasoning_effort` / `message`, same as every other mirror, so there was nothing to drop there.
- **`git diff <base>`** — `SPEC.md` now has a **Base commit** field, recorded at Phase 0 *before any code is written* (`git rev-parse HEAD`, or `none (not a git repo)`). The sweep prompt reads it, and the non-git case hands the reviewer a file list instead of a diff command. Phase 0 says why: after the build there's no way to recover which commit the run started from.
- **`Source: default` vs `deferred_to_author`** — resolved by making it a `Source` value rather than a separate field, so there's nothing to collide: `user` · `default` · `default (deferred_to_author)` · `sweep`. The sweep prompt now keys on a **plain** `default` row in a covered class and explicitly says a `default (deferred_to_author)` row is not an oversight — it was recognised, asked, and handed back. Same correction applied to the Phase 5 report item that named leftover `default` rows.
- **Overlay reference** — removed from both places in the mirror (the header assurance block and the Phase 4 mirror-limitation paragraph). Both now point at the mainline Claude Code skill only.
- **`USER_GUIDE.md`** — deleted. The KV-cache worked example is folded into a `## Worked example` section at the end of `SKILL.md`: the invocation, the `SPEC.md` fields including the base commit, the two Phase-0 rows with one contested `Notes` entry, the `semantic` fork that Phase 3 surfaces, the sweep's actual JSON return, and what the report does and doesn't say. The append/resume claim and the install line are gone with the file.

### Checks

`tests/test_codex_skill_mirror.py` + the argument-hint, skill-groups, reviewer-pins, overlay, corpus-guard, trigger-eval and skill-picker tests: 64 passed, 3 skipped. Full suite: 662 passed, 67 failed — all 67 are the pre-existing `ModuleNotFoundError: No module named 'httpx'` in `test_llm_chat_server` / `test_minimax_chat_server` / `test_llm_chat_review_fallback` / `test_minimax_integration`, identical on `main` and untouched by this branch.

Net: −270 lines.

wanshuiyin:
Merged — thank you, and thank you for turning the trim around the same day; the new version reads as one idea instead of a process manual, and the worked example at the end is exactly the right place for it.

Two small things I fixed in a follow-up commit rather than sending you back for ten lines (151bc19 → follow-up on main):
- The examples still promised `— assume: A-001=…` as an override; with the override string gone from the ledger the skill no longer parses it, so I removed those snippets and left the one-line config edit as the way to reverse a row.
- In the Codex mirror, the submission-gate table said the report unblocks once rows are in the ledger, while the paragraph right under it says a same-family `clean` is only ever provisional. The table now says what the mainline skill and that paragraph say.

It's in the catalog under experiments. Glad to have it.

